PR #3, opened 2018-06-22
Post-London updates

thomas-fossati:
- Remove 1.3 references (based on Hannes's PR #2);
- Protect CID;
- Initial sketch for explicit marking of CID records using ad hoc CTs.

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -210,6 +210,22 @@ protected records. {{new-cid-content-types}} shows the record types to use:
 | heartbeat_with_cid | 28 |
 {: #new-cid-content-types}
 
+# CID authentication
+
+The CID is authenticated.  The MAC of a DTLS record with CID is generated as:
+
+~~~~
+      MAC(MAC_write_key, DTLSCompressed.epoch +
+                            DTLSCompressed.sequence_number +
+                            DTLSCompressed.type +
+                            DTLSCompressed.version +
+                            DTLSCompressed.cid +      // New input
+                            DTLSCompressed.length +
+                            DTLSCompressed.fragment);
+~~~~
+
+   where "+" denotes concatenation.
+
 # Examples
 
 {{dtls-example2}} shows an example exchange where a connection id used
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -1,6 +1,6 @@
 ---
-title: The Datagram Transport Layer Security (DTLS) 1.2 Connection Identifier
-abbrev: DTLS 1.2 Connection ID
+title: The Datagram Transport Layer Security (DTLS) Connection Identifier
+abbrev: DTLS Connection ID
 docname: draft-ietf-tls-dtls-connection-id-latest
 category: std
 updates: 6347
@@ -60,7 +60,7 @@ informative:
 --- abstract
 
 This document specifies the Connection ID construct for the Datagram Transport
-Layer Security (DTLS) protocol version 1.2.  {{I-D.ietf-tls-dtls13}} specifies
+Layer Security (DTLS) protocol.  {{I-D.ietf-tls-dtls13}} specifies
 the Connection ID for DTLS version 1.3.
 
 A Connection ID is an identifier carried in the record layer header that gives the
@@ -91,7 +91,7 @@ these values are insufficient. This is a particular issue in the Internet of Thi
 when devices enter extended sleep periods to increase their battery lifetime. The
 NAT rebinding leads to connection failure, with the resulting cost of a new handshake.
 
-This document defines an extension to DTLS 1.2 to add a connection ID to the
+This document defines an extension to DTLS to add a connection ID to the
 DTLS record layer. The presence of the connection ID is negotiated via a DTLS
 extension.
 
@@ -101,7 +101,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
 document are to be interpreted as described in RFC 2119 {{RFC2119}}.
 
-The reader is assumed to be familiar with DTLS 1.2 {{RFC6347}}.
+The reader is assumed to be familiar with DTLS {{RFC6347}}.
 
 
 # The "connection_id" Extension
@@ -157,16 +157,16 @@ for example by having the length in question be a compile-time constant.
 Note that such implementations must still be able to send other length
 connection identifiers to other parties.
 
-In DTLS 1.2, connection ids are exchanged at the beginning of the DTLS
+In DTLS, connection ids are exchanged at the beginning of the DTLS
 session only. There is no dedicated "connection id update" message
 that allows new connection ids to be established mid-session, because
-DTLS 1.2 in general does not allow TLS 1.3-style post-handshake messages
-that do not themselves begin other handshakes. DTLS 1.2 peers switch to
+DTLS in general does not allow TLS 1.3-style post-handshake messages
+that do not themselves begin other handshakes. DTLS peers switch to
 the new record layer format when encryption is enabled.
 
 # Record Layer Extensions
 
-This extension is applicable for use with DTLS 1.2 and {{dtls-record12}}
+This extension is applicable for use with DTLS and {{dtls-record12}}
 illustrates the record format.
 
 ~~~~
@@ -275,7 +275,7 @@ per-connection to on-path observers. There is no straightforward way to
 address this with connection IDs that contain arbitrary values; implementations
 concerned about this SHOULD refuse to use connection ids.
 
-An on-path adversary, who is able to observe the DTLS 1.2 protocol exchanges between the
+An on-path adversary, who is able to observe the DTLS protocol exchanges between the
 DTLS client and the DTLS server, is able to link the observed payloads to all
 subsequent payloads carrying the same connection id pair (for bi-directional
 communication).  Without multi-homing or mobility, the use of the connection id
@@ -287,7 +287,7 @@ to prevent this, implementations SHOULD attempt to use fresh connection IDs
 whenever they change local addresses or ports (though this is not always
 possible to detect).
 
-This document does not change the security properties of DTLS 1.2 {{RFC6347}}.
+This document does not change the security properties of DTLS {{RFC6347}}.
 It merely provides a more robust mechanism for associating an incoming packet
 with a stored security context.
 
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -171,18 +171,38 @@ This extension is applicable for use with DTLS 1.2 and {{dtls-record12}}
 illustrates the record format.
 
 ~~~~
-  struct {
-     ContentType type;
-     ProtocolVersion version;
-     uint16 epoch;
-     uint48 sequence_number;
-     opaque cid[cid_length];               // New field
-     uint16 length;
-     select (CipherSpec.cipher_type) {
-        case block:  GenericBlockCipher;
-        case aead:   GenericAEADCipher;
-     } fragment;
-  } DTLSCiphertext;
+   struct {
+        ContentType type;
+        ProtocolVersion version;
+        uint16 epoch;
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -1,9 +1,9 @@
 ---
-title: The Datagram Transport Layer Security (DTLS) Connection Identifier
-abbrev: DTLS Connection ID
+title: The Datagram Transport Layer Security (DTLS) 1.2 Connection Identifier
```
ekr: Why 1.2? Can't this be for < 1.3?

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -51,22 +51,25 @@ author:
...
 informative:
   RFC6973:
+  I-D.ietf-tls-dtls13:
+
 --- abstract
 
-This document specifies the "Connection ID" concept for the Datagram Transport Layer Security
-(DTLS) protocol, version 1.2 and version 1.3.
+This document specifies the Connection ID construct for the Datagram Transport
+Layer Security (DTLS) protocol version 1.2.  {{I-D.ietf-tls-dtls13}} specifies
```
ekr: Same comment about 1.2 and below. I tend to think we should just say DTLS before 1.3.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -51,22 +51,25 @@ author:
...
-is accomplished with the help of the 5-tuple. If the source IP address and/or
-source port changes during the lifetime of an ongoing DTLS session then the
-receiver will be unable to locate the correct security context.
+A Connection ID is an identifier carried in the record layer header that gives
+the recipient additional information for selecting the appropriate security
+association.  In "classical" DTLS, selecting a security association of an
+incoming DTLS record is accomplished with the help of the 5-tuple.  If the
+source IP address and/or source port changes during the lifetime of a DTLS
+session then the receiver will be unable to locate the correct security
+context.
```
ekr: Is this jsut reflow?

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -83,26 +86,23 @@ application data protection. Ideally, the second phase where application data is
...
 
-In the current version of DTLS, the IP address and port of the peer is used to
+In the current version of DTLS, the IP address and port of the peer are used to
 identify the DTLS association. Unfortunately, in some cases, such as NAT rebinding,
 these values are insufficient. This is a particular issue in the Internet of Things
-when the device needs to enter extended sleep periods to increase the battery lifetime
-and is therefore subject to rebinding. This leads to connection failure, with the
-resulting cost of a new handshake.
+when devices enter extended sleep periods to increase their battery lifetime. The
+NAT rebinding leads to connection failure, with the resulting cost of a new handshake.
```
ekr, line 92: In future it would be good to have different PRs for off-topic changes.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -132,8 +132,8 @@ the server to use a zero-length CID; the result is the same).
   } ConnectionId;
 ~~~~
 
-A server which is willing to use CIDs will respond with its own
-"connection_id" extension, containing the CID which it wishes the
+A server that is willing to use CIDs will respond with its own
```
ekr: Please don't do this. http://itre.cis.upenn.edu/~myl/languagelog/archives/001464.html

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -162,158 +161,64 @@ connection identifiers to other parties.
...
-: Indicates whether the new CID should be used immediately or is a spare.
-If usage is set to "cid_immediate", then the new CID MUST be used immediately
-for all future records. If it is set to "cid_spare", then either CID MAY
-be used, as described in {{sec-cons}}.
-{:br}
+        ContentType type;
+        ProtocolVersion version;
+        uint16 epoch;
+        uint48 sequence_number;
+        opaque cid[cid_length];               // New field
```
ekr, line 178: I don't think we want to change DTLSPlaintext. This should only be for encrypted records.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -162,158 +161,64 @@ connection identifiers to other parties.
...
-# Examples
+records without knowing how long the Connection ID is.
 
-Below is an example exchange for DTLS 1.3 using a single
-connection id in each direction.
+In order to help the receiver discerning a record bearing a Connection ID from
+one that doesn't, the ContentType of the record containing the cid field MUST
+be: application_data_with_cid(25) instead of application_data(23),
+alert_with_cid(26) instead of alert(24) and heartbeat_with_cid(27) instead of
+heartbeat(24).
```
ekr: "I would write this somewhat differently. In order to allow a receiver to determine whether a record has CID or not, connections which have negotiated this extension use new record types for all protected records. The following table shows the record types to use."

Note you also need handshake. I wonder if anyone cares about heartbeat, though.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -162,158 +161,64 @@ connection identifiers to other parties.
...
-(cid=100)
-                           <========         Application Data
-                                                      (cid=5)
-~~~~
-{: #dtls-example title="Example DTLS 1.3 Exchange with Connection IDs"}
+# Examples
 
-Below is an example exchange for DTLS 1.2 using a connection id used
-uni-directionally from the client to the server.
+{{dtls-example2}} shows an example exchange where a connection id used
```
ekr, line 232: I don't see this example.

ekr (changes requested):

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: I pondered on this as well, then noted that 1.0 is "obsoleted" by 1.2 and decided that we probably should not care about 1.0?  Happy to revert as you suggest if you think it's the sensible thing to do.

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: See reply above.

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: Yep, this must be me inadvertently doing a `gq` in vim.  I'll revert it.

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: ACK

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: ACK

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: I might be horribly wrong, but on a re-handshake are we not dealing with DTLSPlaintext's?

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: ACK, and you're right: I forgot handshake.

(I'd add heartbeat for consistency.)

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: It's the text just below the para.  In the rendered TXT it's "Figure 2".

draft-ietf-tls-dtls-connection-id.md
ekr: I think I would just remove the "1.2" here.

draft-ietf-tls-dtls-connection-id.md
ekr: I don't think we need to explicitly scope it to 1.2.

draft-ietf-tls-dtls-connection-id.md
ekr: I don't follow. We will only send CIDs on encrypted data, and therefore it doesn't need to be in this structure.

draft-ietf-tls-dtls-connection-id.md
ekr: OK.

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: Yes, I get that.  The thing that confuses me then is how to interpret this bit of [RFC5246](https://tools.ietf.org/html/rfc5246#section-7.4) regarding (re-)handshake records:
```
   [...] Handshake messages are supplied to
   the TLS record layer, where they are encapsulated within one or more
   TLSPlaintext structures, which are processed and transmitted as
   specified by the current active session state.
```
And this other bit of [RFC5246](https://tools.ietf.org/html/rfc5246#section-6.2.3):
```
    The encryption and MAC functions translate a TLSCompressed
    structure into a TLSCiphertext.
```
Which is the reason why I originally added CID to both `DTLSPlaintext` and `DTLSCompressed`.

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: OK, that makes sense.  DTLS 1.3 will then supersede this mechanism with its own CID machinery.

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: ACK

draft-ietf-tls-dtls-connection-id.md
ekr: Sure, and the way to think about this is that you are adding the CID as part of that transformation.

thomas-fossati: Got it, thanks.

(We still need to say that CID is fed into the MAC algorithm.)

thomas-fossati: > (We still need to say that CID is fed into the MAC algorithm.)

Done in 0b40194

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -159,161 +157,79 @@ for example by having the length in question be a compile-time constant.
...
-                                                      (cid=5)
-                                                     Finished
-                                                      (cid=5)
+~~~~
+      MAC(MAC_write_key, DTLSCompressed.epoch +
+                            DTLSCompressed.sequence_number +
+                            DTLSCompressed.type +
+                            DTLSCompressed.version +
+                            DTLSCompressed.cid +      // New input
+                            DTLSCompressed.length +
```
hannestschofenig, line 224: Do we need to modify the key derivation function to include the CID? Why do you think we need it?

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: Did we conclude that the CID only needs to be added to the DTLSCiphertext (and not to the DTLSPlaintext and the DTLSCompressed)?

ekr:
I also do not understand that

On Mon, Jun 25, 2018 at 8:01 AM, Hannes Tschofenig <notifications@github.com

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: I got confused by this comment from the 101 minutes:
```
Subodh: Re-read draft and the TLS conn-ID isn't under the authentication?

ekr: Need to write it down in 1.2.
```
which I thought it meant we forgot to put CID in under authentication.

If this is not the case, I can remove the modified MAC computation together with the DTLSCompressed stuff, which was just needed to reference DTLSCompressed.cid in the algorithm.

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: (This the MAC computation, not the key derivation. Though, this is probably not needed -- see my reply above.)

ekr:
Sorry, you're right. It needs to be in the MAC, but not the key derivation

-Ekr

hannestschofenig:
I don't think we need to include the CID in the MAC since the CID is just the key name. In regular DTLS the 5 tuple is the key name and we don't include it in the MAC calculation either. 

If someone fakes a packet and places a wrong CID in there then they MAC will just fail since either (a) the wrong key will be looked up or (b) no key is found at all. In either case there is no issue.

PR #13, opened 2018-10-21
Use a wrapped content type as proposed by Martin Thomson and

ekr:
discussed in YUL.

@martinthomson

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -211,12 +211,15 @@ encrypted.
 
 ~~~~
      struct {
-         opaque content[TLSPlaintext.length];
+         opaque compressed[TLSCompressed.length];
          ContentType type;
          uint8 zeros[length_of_padding];
       } DTLSWrappedCompressed;
 ~~~~
 
+
+compressed
+: The value of DTLSCompressed.fragment
 
 type
 : The true content type.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -193,21 +194,38 @@ Note that for both record formats, it is not possible to parse the
...
 
-The CID value, when present, is included in the MAC calculation for the DTLS
+When CID is being used, the DTLSCompressed value is first wrapped
+along with the true content type and padding into a DTLSWrappedCompressed
+value prior to encryption. The DTLSWrappedCompressed value is then
+encrypted.
+
+~~~~
+     struct {
+         opaque content[TLSPlaintext.length];
```
chris-wood: Do we need a description of `content` in the list below?

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -193,21 +194,38 @@ Note that for both record formats, it is not possible to parse the
...
+
+
+type
+: The true content type.
+
+zeroes
+: Padding, as defined in {{RFC8446}}.
+{:br}
+
+In addition, the CID value is included in the MAC calculation for the DTLS
```
chris-wood: Is this just a clarifying sentence? The expression below suggests it was already included in the MAC computation.

PR #19, opened 2018-11-20
Remove exposition

martinthomson:
> This is effectively the simplest possible design that will work.

Not that important to say.

> Previous design ideas for using cryptographically generated session ids, either using hash chains or public key encryption, were dismissed due to their inefficient designs. 

Same.

> Note that a client always has the chance to fall back to a full handshake or more precisely to a handshake that uses session resumption.

Why this is a relevant statement is baffling.  The point that is being made is that more complex designs might allow a connection to continue in the face of path changes (and NAT rebindings), but these are far too complex and in cases where a connection ID is insufficient, a new connection is probably easier to manage.

But you don't need to say all of that.

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -140,6 +140,9 @@ indicates that the server will send with the client's CID but does not
 wish the client to use a CID (or again, alternately, to use a
 zero-length CID).
 
+Having endpoints select the CID that is used by peers gives implementations
+flexibility in the choice of how to format and generate CIDs.
+
 When a session is resumed, the "connection_id" extension is
 negotiated afresh, not retained from previous connections in
 the session.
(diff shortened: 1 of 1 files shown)


ekr:
I'd be willing to revise this a bit, but I think it's important to recognize that we are aware of the shortcomings of this design but that it was chosen for a reason. If you want to expand on the last sentence, I'd be fine with that, otherwise I prefer to leave as-is.

hannestschofenig:
I also think it is valuable for the reader to explain that there are shortcomings of this design and, since we have a slightly different design in DTLS 1.3, that there is a possibility to get better privacy protection when moving to DTLS 1.3. For this reason I would not remove the sentence or the paragraph.

hannestschofenig:
I created an alternative proposal in PR #41. 
What do you think, Martin?

martinthomson:
#41 is better.

Issue #28, opened 2018-12-11
Should mixed usage of cid and no-cid for one connection-endpoint be supported? 


boaks:
If a cid is negotiated for a connection-endpoint, is it then intended to use that in all records for that connection-endpoint (at least all first records of a datagram)?

hannestschofenig:
Yes, if you negotiate it with the other peer then you use it. Note, however, that there is still the option that the client uses a CID in the records sent to the server while the server doesn't include a CID in the records sent to the client. This is what is shown in the example. IMHO is a very common case.

boaks:
My intention was, that, even if y cid is negotiated, and a "starting record" contains such a cid, following records may use the old record layer to save some bytes. Especially, if the cid could be extended into the not encrypted handshake messages, I would prefer to just use the cid for the first record in the datagram and save the bytes by not using the cid in the other handshake messages of the same datagram.

hannestschofenig:
The mix of CID and non-CID payloads should be supported by the current draft since (I believe) we don't force the client or the server to always attach the CID. I double-checked the text and I didn't see anything that would prohibit you doing so. Of course, we should be explicit about features we offer. 

FWIW we allow this behavior (and more record format changes) in DTLS 1.3 thanks to the bitmask in the record header and the variable length encoding.

boaks:
Thanks, that clarification will help to implement it proper.
Just one hint for the spec:
Using CID and non-CID, it must be clear, that potentially messages are reordered and so non-CID messages may be silently ignored. Therefore, mixing CID and non-CID is possible, but must be used with special care.

hannestschofenig:
I am not sure why this matters here since messages may in general get re-ordered with a UDP-based transport

boaks:
Yes, and one consequence of this reorder would be, that a message without CID send after a message with CID, are received potentially in the reverse order, what then cause **additionally** a drop. It's also true, the messages may always get dropped. I just want to make it clear by the hint, that such mix must be used with special care.

ekr:
I believe you should be required to send CID always. That's the implied semantics of the extension and there's no basis for knowing that the receiver would like something different.

boaks:
> there's no basis for knowing that the receiver would like something different

This draft would be the basis, because it would define, what the receiver should know :-). 

[...]

Issue #25, opened 2018-11-24
Record Layer Protection - MAC


boaks:
https://github.com/tlswg/dtls-conn-id/blob/master/draft-ietf-tls-dtls-connection-id.md#record-payload-protection

According RFC5246, the record layer protection is split into 3 sections
- NULL and stream (stream is not used for DTLS)
- CBC
- AEAD

in this draft, only the NULL is referenced
https://tools.ietf.org/html/rfc5246#section-6.2.3.1

If the CID value should be included in that MAC, I would expect, that this draft specifies how to do this for all three variants. By the way, I'm not sure, why the CID should be explicitly included. The CID "points" to the "keys" used to calculate the MAC, with a wrong CID, the wrong "key" will be used and that should cause a MAC violation anyway.

ekr:
I can fix this text, but there are multiple CIDs pointing to the same keys, so we need to have the CID in the MAC

boaks:
Sorry, I didn't get that. Why are "multiple CIDs" pointing to the same keys? 
Is "keys" referring to https://tools.ietf.org/html/rfc5246#section-6.1 

```
client write MAC key
server write MAC key
client write encryption key
server write encryption key
client write IV
server write IV
```

may be I miss something, if ip-address:port is used to identify the "keys", the MAC ensures, that the same keys are used. But the MAC itself is not based on the ip-address:port. or I'm wrong? If, I didn't miss something, then the cid is not required to be directly included in the MAC, as the ip-address:port. Instead is indirect validated with the MAC, as the ip-address:port would also be validated.

ekr:
You can change CIDs during the connection. Thus, if it's not included in the MAC, you can substitute CIDs.

boaks:
Sorry, I still can't see the point.
Do you mean, the CID could be exchanged during the transmission?
FMPOV this seems to be no issue, which could be fixed by including it explicit in the MAC.
The ip-address:port may also be exchanged, but are not included explicit in the MAC (this would fail on every NAT).  

Receiver side:
```
<CID1> => <mac keys1, enc keys1> 
<CID2> => <mac keys2, enc keys2> 
<CIDy> => no keys
``` 

Record `<CID1>` contents get protected by a MAC calculated with `<mac keys1>`.
If the `<CID1>` is exchanged during transmission with `<CIDy>`, this fails, because, no keys are mapped at the receiver.
[...]

Issue #22, opened 2018-11-20
cid=empty and MAC calculation


martinthomson:
If this feature is negotiated on, then the MAC calculation changes, even if no connection ID is sent, because of `cid_length`.  This needs to be clearly sign-posted.

boaks:
My preferred interpretation is:

`cid=empty` uses the original RFC6347, section 4.1 record format

https://tools.ietf.org/html/rfc6347#section-4.1

indicated by the already used content type ids. That includes also the unchanged MAC.

hannestschofenig:
Created a pull request to add clarifying text: https://github.com/tlswg/dtls-conn-id/pull/35

boaks:
@martinthomson 

it would be great, if you could explain, if you want to use the new record type with an empty CID in order to obfuscate the DTLS payload length. Please note, that currently the specification of this draft turns more into a "strict" usage (see issue #28) which will not support the new record type with an empty CID.

martinthomson:
Interesting question.  Yes, I would like to be able to use the new record type to pad records, but I don't think that it is an especially important consideration.  If the design is easier without that, and I suspect that it is, then I am OK with not doing that.  A one-octet connection ID is not that much of a burden to pay to get this feature.

boaks:
> A one-octet connection ID is not that much of a burden to pay to get this feature.

So you would then got for adding a 1 byte cid, but still use the ip-address:port to map it to the security context (keying stuff)? Or do you consider only a few connections?

For me that is just one of the scenario's, I'm really afraid of! 
It makes the usage of cid more indeterministic, if the spec "ignores" that kind of usage instead of explicitly considering it!

FMPOV, it will be easier, to support then a empty CID record indicating, that still the ip-address:port must be used instead of a 1 byte cid, but then being not strict on the mapping.

hannestschofenig:
Addressed in submitted version -04

PR #29, opened 2018-12-27
 Updated Record Layer section

hannestschofenig:


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -229,22 +229,41 @@ zeroes
...
-                            DTLSWrappedCompressed.length +
-                            DTLSWrappedCompressed.fragment);
-   where "+" denotes concatenation.
+    additional_data = seq_num + DTLSWrappedCompressed.type + 
+	                  DTLSCompressed.version + DTLSCompressed.length + 
+					  cid + cid_length + cid_codepoint + 
+					  DTLSWrappedCompressed.length + 
+					  DTLSWrappedCompressed.fragment;
+					  
+    where "+" denotes concatenation.
```
boaks, line 274: Why should `DTLSWrappedCompressed.type` be included, it's contained in the encrypted part?
Why both, `DTLSCompressed.length` and `DTLSWrappedCompressed.length`, should be included?
(By the way, I don't know, what  `DTLSWrappedCompressed.length` is. It's not defined in the `DTLSWrappedCompressed` struct.)

hannestschofenig: I modified the list of fields in the additional data calculation.

boaks: The tls `seq_num` of https://tools.ietf.org/html/rfc5246#section-6.2.3.3 is in my opinion not the dtls `seq_num` of https://tools.ietf.org/html/rfc6347#section-4.1 ! Please don't remove the `epoch` field.

boaks: Uups, sorry forget the upper comment about the epoch! I read the last commit in isolation ;-(.

thomas-fossati: Thanks Hannes, looks good to me.

PS: not that it makes any difference, but adding the fields in the order in which they appear in the header
```
   seq_num + type + version + cid_length + cid + length
```
looks slightly more consistent with the previously established convention?

thomas-fossati: Another small note: I think we should not remove the MAC computation.  I understand we want to move to an AEAD-only world, but this needs to work with every mode supported in 1.2...

boaks: > PS: not that it makes any difference, but adding the fields in the order in which they appear in the header

There is no field `cid_length` in the header :-).

hannestschofenig: While there is no such field in the header that doesn't mean we cannot include it in the additional data calculation.

hannestschofenig: Thomas, if you can add the text for the other MAC calculations then that's fine with me. I only care about the AEAD ciphers since we recommend them in the DTLS / TLS IoT profiles draft.

boaks: > While there is no such field in the header that doesn't mean we cannot include it in the additional data calculation.

Sure, but why should it then be added?

boaks: Which size is considered for the `cid_length`? The same as in the extension (`opaque cid<0..2^8-1>`, so 1 byte as `uint8`)?

hannestschofenig: It would be 1 byte length but your remark is fair. IMHO one would have to do a formal analysis to say whether it really has to be added or not. Currently, we are just adding fields to the additional data / MAC calculation without much justification

boaks: OK, thanks! That helps me to understand the current definition.

hannestschofenig: I updated the additional_data calculation again. I removed the cid_length since it is only implicit and I have re-ordered the fields as they appear on the wire.

hannestschofenig: Corrected the order again since the seq_nr was used as the initial value in RFC 7366, even though this is not how it is represented on the wire.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -255,25 +255,37 @@ Other fields are defined in RFC 6347. Note that this specification does
...
+: This value contains the outer-header content type, i.e. the tls12_cid. 
+
+version
+: This value contains the version number. 
+
+length
+: This value contains the length information in the outer-header. 
 
+cid
+: Value of the connection id.
```
thomas-fossati: I'd rather say "Value of the *negotiated* connection id" to distinguish it from what is sent on the wire.  And it's maybe worth adding something like "Note that this is empty in case a 0-length CID has been negotiated".

I'd also add an entry for cid_length here, specifying that even if a 0-length CID has been negotiated, and therefore no CID appears on the wire, a cid_length of 0 MUST be added to the AD.

hannestschofenig: I added that text

jsalowey (approved):
These changes look appropriate.

draft-ietf-tls-dtls-connection-id.md
boaks: Sorry, if the usage of "empty CID" with the new record type is intended to obfuscate the payload length, this is clearly a new use-case and obvious no NAT solution!

So, please add this new use-case explicit to the introduction!

And please provide an example, which shows that adding the cid_length make sense!
In other discussion "empty CID in the new record types" are denied.

boaks:
> These changes look appropriate.

According https://github.com/tlswg/dtls-conn-id/issues/28#issuecomment-458032109

> As I said, empty CID should mean "use the 6347 framing"

this is obvious far away from "appropriate".

I would appreciate, if the scope of this draft could be clarified, before adding additional features.

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: There is no new "payload length obfuscation" use case that is enabled by this change.

I think you are confusing on-the-wire framing with how authentication data is computed.

Hannes's change here is only about MAC computation.  The gist is that 0-length CIDs have 6347 framing but their MAC is computed slightly differently from a "normal" 6347 record.

boaks: Hm, so want you to ensure, that if a empty CID is used in the Hello Extension, and therefore a RFC 6347 record is used later, the MAC is different compared to a RFC 6347 record, which is used without using a empty cid in the Hello Extension?

Why? What would be the advantage? What are the examples, which show this advantage?

FMPOV: the MAC includes already much keying material, the cid_length willl hardly improve that. From the discussion in #25 two arguments for including the MAC have been named: 

https://github.com/tlswg/dtls-conn-id/issues/25#issuecomment-444890710

"Yes, but it might be true with some extension in the future. Moreover, authenticating the whole header is just good practice."

I simply just don't see the justification, to complicate RFC6347 MAC calculations.

boaks: > There is no new "payload length obfuscation" use case that is enabled by this change.

In my opinion, this https://github.com/tlswg/dtls-conn-id/issues/22#issuecomment-458758785 indicates, that it is a use-case :-).

Therefore I started #40 .

hannestschofenig:
The issue you raised in https://github.com/tlswg/dtls-conn-id/issues/28#issuecomment-458032109 is important and will be addressed in a separate PR. I would like to merge some of the PRs to that we do not have too many conflicts. This PR only provides the text for the additional_data and, since we have changed our mind about it so many times already, I would like to get it into the working draft. Then, Thomas can make a contribution for the MAC calculation for non-AEAD ciphers.

boaks:
@hannestschofenig 

Thank you very much for great patience! 
It's sure still easy to implement, even with a new MAC for the old RFC 6347 record. 
As I tried to point out, if there are new use-cases, they should not be obfuscated by introducing only "maverick fields" somewhere, in my opinion, they should be addressed clearly. 

I still see no argument to include the cid_length into the MAC.

PR #30, opened 2018-12-27
Updated example

hannestschofenig:


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -249,45 +249,63 @@ as follows:
...
+<cid=100>                   
+
+                                           [ChangeCipherSpec]
+                            <--------                Finished
+
+
+Application Data**          ========>
+<cid=100>
+
+                            <========        Application Data
```
boaks, line 291: Please add a comment to clarify issue #22.
Should the new record layer be used (https://github.com/tlswg/dtls-conn-id/blob/master/draft-ietf-tls-dtls-connection-id.md#record-layer-extensions), even if the cid is empty? Or should the RFC6347 record layer be used?

boaks: The empty cid is done in PR #35. Thanks!

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -249,45 +249,63 @@ as follows:
...
-{: #dtls-example2 title="Example DTLS 1.2 Exchange with Connection IDs"}
+{: #dtls-example2 title="Example DTLS 1.2 Exchange with Connection ID"}
+
+Note: In the example exchange the CID is included in the record layer 
+once encryption is enabled. In DTLS 1.2 only one handshake message is 
+encrypted, namely the Finished message. Since the example shows how to 
+use the CID for payloads sent from the client to the server only the 
+record layer payload containing the Finished message (marked with *) 
+contains a CID. Application data payloads sent from the client to the 
+server all contain a CID, as shown in **.
```
ekr: This footnoting seems confusing. Why not just mark messages which have a CID with the CID.

PR #32, opened 2019-01-07
Minor editorial changes

hannestschofenig:


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -194,7 +196,7 @@ Note that for both record formats, it is not possible to parse the
 records without knowing how long the Connection ID is.
 
 In order to allow a receiver to determine whether a record has CID or not,
-connections which have negotiated this extension use a distinguished
+connections, which have negotiated this extension, use a distinguished
```
ekr: These commas are actually wrong. Commas indicate a nonrestrictive clause, but this is restrictive.

ekr (approved):

Issue #23, opened 2018-11-20
cid, CID, connection id, and connection ID


martinthomson:
It probably makes sense to be a little more consistent.

hannestschofenig:
Makes sense. I will do this after the current wave of pull requests have been reviewed and accepted/rejected.

hannestschofenig:
I addressed this issue in PR #42

hannestschofenig:
Addressed with submitted draft version -04

PR #35, opened 2019-01-07
Record Layer Clarification

hannestschofenig:
Concerning the use the connection_id=empty

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -103,7 +103,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
 document are to be interpreted as described in RFC 2119 {{RFC2119}}.
 
-The reader is assumed to be familiar with DTLS {{RFC6347}}.
+The reader is assumed to be familiar with DTLS 1.2 {{RFC6347}}.
 
 
 # The "connection_id" Extension
@@ -139,6 +139,14 @@ client to use when sending messages towards it. A zero-length value
 indicates that the server will send with the client's CID but does not
 wish the client to use a CID (or again, alternately, to use a
 zero-length CID).
+
+Whenever a zero-length CID has been negotiated then the RFC 6347-defined 
+record format MUST be used (see Section 4.1 of {{RFC6347}}). For example, 
+in {{dtls-example2}} the use of a connection ID has been successfully 
+negotiated between a client and a server whereby the client uses 
+the record format defined in this specification to transmit the CID value 
+(100) to the server while the server still uses the RFC 6347-defined 
+record format without a CID when transmitting a record to the client. 
 
 When a session is resumed, the "connection_id" extension is
 negotiated afresh, not retained from previous connections in


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -140,6 +140,14 @@ indicates that the server will send with the client's CID but does not
 wish the client to use a CID (or again, alternately, to use a
 zero-length CID).
 
+Whenever a zero-length CID has been negotiated then the RFC 6347-defined
```
martinthomson, line 143: I think that you need to explicitly mention the MAC calculation in this paragraph.  A zero-length CID will naturally look the same on the wire, it's how the MAC is calculated is the real trap here.

boaks: > A zero-length CID will naturally look the same on the wire

I didn't get it. A zero-length CID uses the RFC6347 record layer and so the types used are defined in https://tools.ietf.org/html/rfc5246#section-6.2.1 

```
enum {
change_cipher_spec(20), alert(21), handshake(22),
application_data(23), (255)
} ContentType;
```

if the cid is none-empty, the `type` is the new one `tls12_cid(25)`.

So I'm not sure, do you have a use-case, which makes a difference using a empty cid or no cid? I can't see the benefit in distinguish that.

ekr: ISTM that the right way to interpret this is:

- The type field indicates whether CID is present
- You then use the CID and/or address information to look up the connection
- If the CID presence/absence doesn't match what was negotiated, that's a MAC error
- You then use the MAC algorithm associated with expected presence or absence.

This resolves two contentions points:

1. Ambiguity about the MAC
2. Mixed CID and non-CID packets.

jsalowey: This looks like a good way forward,  can we clarify this in the PR?

boaks: Would it be possible, that the use of "empty CIDs" could be addressed explicit?

My impression is, before introducing the `uint8 zeros[length_of_padding] ` an empty CID in the extension was used by a client to indicate, that the client will be able to accept a CID from the server. It was not to force the server to use the new record layer when sending data to the client even with a empty CID.

With the introduction of the new record layer with the `uint8 zeros[length_of_padding] `, this draft has a additional functionality, it could be used to obfuscate the payload length. So, if this extension is not only intended to solve the address (NAT) issue, and it is intended to offer such a "payload length  obfuscation", then that should be made more explicit.

boaks: > If the CID presence/absence doesn't match what was negotiated, that's a MAC error

I'm not sure about that. If in "Absence" a RFC6347 record is used, and it matches the right connection with the address:port, this will not cause a MAC error (it's a valid RFC6347 MAC). instead, if the "cid requires cid" (see #28) should be implemented and a packet with a record without CID MUST be dropped, that will be an (easy) additional check.

boaks: > If the CID presence/absence doesn't match what was negotiated, that's a MAC error

FMPOV, that would only be true, if https://github.com/tlswg/dtls-conn-id/blob/master/draft-ietf-tls-dtls-connection-id.md#record-payload-protection applies for **all** records, not only tls_cid.
 
> Moreover, authenticating the whole header is just good practice.

If the new MAC is applied independent from the record type, the cid-values are not from the header.

But I would prefer a interpretation, where the new MAC is bound to the new record type.

boaks: > If the CID presence/absence doesn't match what was negotiated, that's a MAC error

Or is the intention to **assume**, that's a MAC error (and be processed as such a MAC error). But it's not an error caused by the calculation of the MAC, because the MAC of a RFC6347 record would be valid for itself.

ekr:
On Mon, Jan 28, 2019 at 12:45 AM Achim Kraus <notifications@github.com>
wrote:

> *@boaks* commented on this pull request.
> ------------------------------
>
> In draft-ietf-tls-dtls-connection-id.md
> <https://github.com/tlswg/dtls-conn-id/pull/35#discussion_r251320406>:
>
> > @@ -140,6 +140,14 @@ indicates that the server will send with the client's CID but does not
>  wish the client to use a CID (or again, alternately, to use a
>  zero-length CID).
>
> +Whenever a zero-length CID has been negotiated then the RFC 6347-defined
>
> If the CID presence/absence doesn't match what was negotiated, that's a
> MAC error
>
> I'm not sure about that. If in "Absence" a RFC6347 record is used, and it
> matches the right connection with the address:port, this will not cause a
> MAC error (it's a valid RFC6347 MAC). instead, if the "cid requires cid"
> (see #28 <https://github.com/tlswg/dtls-conn-id/issues/28>) should be
> implemented and a packet with a record without CID MUST be dropped, that
> will be an (easy) additional check.
>

MAC error is the generic failure for "malformed packet"

boaks:
OK, so it's **assumed** to be an MAC error and processed as that.

hannestschofenig:
In a discussion with Achim we came to a conclusion on the use of CIDs for handshake messages, see https://github.com/tlswg/dtls-conn-id/issues/23

Regarding the MAC calculation there are two design approaches: 

Approach I: 
  - If datagram contains new content type then use new MAC calculation.
  - if datagram uses regular content type then use previous MAC calculation. 

Approach II: 
  - If a CID was negotiated (regardless whether it is actually used by both parties) then the new MAC calculation is used (irrespectively of what content type is used in the datagram).

What do you prefer?

boaks:
I would prefer approach I.

For approach II I would appreciate, If the arguments would be explicitly expressed and examples are provided, which shows the positive effect.

ekr:
On Fri, Feb 1, 2019 at 2:29 AM Hannes Tschofenig <notifications@github.com>
wrote:

> In a discussion with Achim we came to a conclusion on the use of CIDs for
> handshake messages, see #23
> <https://github.com/tlswg/dtls-conn-id/issues/23>
>
> Regarding the MAC calculation there are two design approaches:
>
> Approach I:
>
>    - If datagram contains new content type then use new MAC calculation.
>    - if datagram uses regular content type then use previous MAC
>    calculation.
>
>
I agree with this, except that it should be an error to use a content type
incompatible with the CID negotiated.

boaks:
> I agree with this, except that it should be an error to use a content type incompatible with the CID negotiated.

Though the discussion details are here

https://github.com/tlswg/dtls-conn-id/pull/29#discussion_r252569064

the part 

> if datagram uses regular content type then use previous MAC calculation.

is used for empty CIDs.

boaks:
#43 (case 2 there) may indicate "alternative III":

If the peers negotiate to use CID, all datagram contains new content type and use the new MAC calculation, even if a empty CID is negotiated. If a empty CID is used, naturally the address:port is used to lookup the connection.

Issue #43, opened 2019-02-02
cid=empty=?padding?


martinthomson:
>A zero-length value indicates that the client is prepared to send with a connection ID but does not wish the server to use one when sending (alternately, this can be interpreted as the client wishes the server to use a zero-length CID; the result is the same).

This implies that an empty CID and an absent CID are identical.  But they don't have to be. If both endpoints negotiate a zero- length CID, then this extension could be used to enable record padding.

The doc is not clear on this point. If a peer sends an empty connection_id extension, what are you supposed to do:

1. MUST send records in the old format.
2. MUST send tls12_cid records (which could be padded).
3. MAY send either.

It seems like option 2 or 3 give us a nice new feature - the ability to pad records - though 2 forces an additional overhead of 1 byte. Option 3 is somewhat less deterministic though (it doesn't seem like it would be harder to implement though).

ekr:
On Fri, Feb 1, 2019 at 9:38 PM Martin Thomson <notifications@github.com>
wrote:

> A zero-length value indicates that the client is prepared to send with a
> connection ID but does not wish the server to use one when sending
> (alternately, this can be interpreted as the client wishes the server to
> use a zero-length CID; the result is the same).
>
> This implies that an empty CID and an absent CID are identical. But they
> don't have to be. If both endpoints negotiate a zero- length CID, then this
> extension could be used to enable record padding.
>
> The doc is not clear on this point. If a peer sends an empty connection_id
> extension, what are you supposed to do:
>
>    1. MUST send records in the old format.
>    2. MUST send tls12_cid records (which could be padded).
>    3. MAY send either.
>
> It seems like option 2 or 3 give us a nice new feature - the ability to
> pad records - though 2 forces an additional overhead of 1 byte. Option 3 is
> somewhat less deterministic though (it doesn't seem like it would be harder
> to implement though).
>
I agree that the old text is wrong. It comes from a previous encoding and
MAC structure. I don't care about the padding feature for (2) or (3), as
this is a point fix for DTLS 1.2. I could live with (2) but the
nondeterminism in (3) seems bad.

hannestschofenig:
Addressed in submitted version -04

Issue #40, opened 2019-01-29
Scope of this draft - only NAT traversing or "more"?


boaks:
During the discussion of the last weeks, my impression of the scope of this draft has changed.

At the begin it was clearly the solution for issues caused by a NAT.

With the questions from others about the usage of a empty CID in the MAC, I started to wonder, why this should be used. In the meantime in think, using the new record type (even with an empty CID) would enable peers to obfuscate the DTLS payload length. That feature may be even used without a real CID (therefore using a empty CID). I'm not sure, if this is really planned. I asked for it, but unfortunately, didn't receive an answer. If that usage is planned, I think it should be addressed explicitly.  
See:
Issue #22
https://github.com/tlswg/dtls-conn-id/pull/35#discussion_r245795214

And working on the implementation, I got also the idea, to use this extension to improve the protection of the handshake itself (see #26). Currently the protection seems to be mainly the "randomness of the ip-address:port". But this randomness is not really defined and so using a random CID to lock the handshake may help. The interesting point of this usage would be, that it get's more important, when the address doesn't change (less randomness), so it targets the opposite environment as the original idea (frequently changing (by NAT) to static reachable). In the end this may end up in a "large CID used in the handshake" and no need to use the CID after the handshake. 

With these other "use-cases", I'm not sure, if the "strict" definitions in #28 should be implemented or if the definitions should be "relaxed".
The worst case would be, that the "strict" is mostly implemented, but the users also want the other scenarios independent for the "original idea".

boaks:
See https://github.com/tlswg/dtls-conn-id/issues/26#issuecomment-459676265

Issue #27, opened 2018-12-10
Parse the records without knowing how long the Connection ID is


boaks:
https://github.com/tlswg/dtls-conn-id/blob/master/draft-ietf-tls-dtls-connection-id.md#record-layer-extensions

> Note that for both record formats, it is not possible to parse the records without knowing how long the Connection ID is.

FMPOV,  searching for the "length field" assuming that it's that uint16 pointing to the end of the UDP datagram, should result in many cases in the right length for the connection id. Moving the connection id after the "length field", will hide it from that search, but with the AEAD explicit nonce copied from the header, it will also be easy to find there.
Therefore, if it's intended to hide the connection id length in tls12_cid records, it must be after the "length field", and the explicit nonce could not longer be copied from the header to the fragment (saving 8 bytes).
It's no issue, if there is no intention to hide the connection id length. But then the misguided note above should be removed.

ekr:
We're not hiding the CID length. It's transmitted in the clear in the handshake.

boaks:
So what's the message of  

> Note that for both record formats, it is not possible to parse the records without knowing how long the Connection ID is.

it is in many cases possible.

Issue #26, opened 2018-11-24
CID usage during handshake


boaks:
https://github.com/tlswg/dtls-conn-id/blob/master/draft-ietf-tls-dtls-connection-id.md#examples

The example shows in flight 5

Certificate                 -------->
ClientKeyExchange
CertificateVerify
[ChangeCipherSpec]
Finished
(cid=100)

Does this imply that the flight 5 is send with cid = 100 or only the Finished?

I would prefer to be able to send also the other handshake messages of the flight with a cid, that would make it harder to disturb handshakes by spoofed handshake messages.

hannestschofenig:
In DTLS 1.2 only the Finished message is encrypted. In the current design we make use of the CIDs only once encryption is enabled. I should explain that only the Finished message contains a CID here.

hannestschofenig:
See additional explanation in PR #30: https://github.com/tlswg/dtls-conn-id/pull/30

boaks:
I would enjoy it, if the cid could be extended into the not encrypted handshake messages also. Is there any reason, not to do so?

My consideration is to make the handshake "slightly more" protected against spoofing.
The current protection in RFC6347 is limited to https://tools.ietf.org/html/rfc6347#section-4.2.1 and https://tools.ietf.org/html/rfc6347#section-4.1.2 .
The first applies only to the CLIENT_HELLO ("cookie" protection again spoofed CLIENT_HELLOs). The second (MAC and Anti-Replay) is only usable after a successful handshake.  If an attacker spoofs other handshake messages, the attacker may block a handshake from being successful (even if the time window for that is very small and so it would be a rather expensive attack). Using a CID would enable a implementation either to handshake even when addresses are changing during the handshake or to protect the handshake slightly more (not supporting address changes but) use the CID to validate the handshake message belonging to this handshake. I would leave the chosen strategy to the implementation, enabling implementations to flexible chose the strategy depending on the current handshake load (means: supporting address changes in handshakes as normal operation, but switch dynamically to the more protective one, if a certain relative number of handshake fails.)  

My current work in progress:
https://github.com/eclipse/californium/archive/wip_add_dtls_cid.zip
  
contains already this feature as experimental extension. It's easy to be added to a cid implementation.

hannestschofenig:
[...]

PR #44, opened 2019-02-07
Rewrite of the record layer and MAC text

hannestschofenig:
Text to address the discussion in https://github.com/tlswg/dtls-conn-id/pull/35

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -192,7 +192,7 @@ connections which have negotiated this extension use a distinguished
 record type tls12_cid(25). Use of this content type has the following
 three implications:
 
-- The CID field is present.
+- The CID field is present and contains one or more bytes.
 - The MAC calculation follows the process described in {{mac}}.
 - The true content type is inside the encryption envelope, as described
   below.
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -97,7 +97,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
 document are to be interpreted as described in RFC 2119 {{RFC2119}}.
 
-The reader is assumed to be familiar with DTLS 1.2 {{RFC6347}}.
+This document assumes familiarity with DTLS 1.2 {{RFC6347}}.
 
 # The "connection_id" Extension
 
@@ -159,15 +159,15 @@ the ServerHello messages then the peers need to take the following steps.
 
 The DTLS peers determine whether incoming and outgoing messages need 
 to use the new record format, i.e., the record format containing the CID. 
-The new record format and the CID content type is only used once encryption 
+The new record format with the the tls12_cid content type is only used once encryption 
 is enabled. Plaintext payloads never use the new record type and the CID content 
 type. 
 
 For sending, if a zero-length CID has been negotiated then the RFC 6347-defined 
 record format and content type MUST be used (see Section 4.1 of {{RFC6347}})
-else the new record layer format and the new content type defined in {{dtls-record12}} MUST be used. 
+else the new record layer format with the tls12_cid content type defined in {{dtls-record12}} MUST be used. 
 
-When transmitting a datagram with the new record format and the new content type, 
+When transmitting a datagram with the tls12_cid content type, 
 the new MAC computation defined in {{mac}} MUST be used.
 
 For receiving, if the tls12_cid content type is set, then the CID is used to look up 
@@ -176,7 +176,7 @@ then the connection and security association is looked up by the 5-tuple and a
 check MUST be made to determine whether the expected CID value is indeed 
 zero length. If the check fails, then the datagram MUST be dropped. 
 
-When receiving a datagram with the new record format and the new content type, 
+When receiving a datagram with the tls12_cid content type, 
 the new MAC computation defined in {{mac}} MUST be used. When receiving a datagram
 with the RFC 6347-defined record format the MAC calculation defined in Section 4.1.2 
 of {{RFC6347}} (and Section 4.1.2.4  of {{RFC6347} for use with AEAD ciphers) MUST 
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -165,16 +165,16 @@ type.
 
 For sending, if a zero-length CID has been negotiated then the RFC 6347-defined 
 record format and content type MUST be used (see Section 4.1 of {{RFC6347}})
-else the new record layer format and the new content type MUST be used. 
+else the new record layer format and the new content type defined in {{dtls-record12}} MUST be used. 
 
 When transmitting a datagram with the new record format and the new content type, 
 the new MAC computation defined in {{mac}} MUST be used.
 
-For receiving, if the CID content type is set, then the CID is used to look up 
-the connection and the security association. If the CID content type is not set, 
+For receiving, if the tls12_cid content type is set, then the CID is used to look up 
+the connection and the security association. If the tls12_cid content type is not set, 
 then the connection and security association is looked up by the 5-tuple and a 
-verification MUST be made to determine whether the expected CID value is indeed 
-zero length.
+check MUST be made to determine whether the expected CID value is indeed 
+zero length. If the check fails, then the datagram MUST be dropped. 
 
 When receiving a datagram with the new record format and the new content type, 
 the new MAC computation defined in {{mac}} MUST be used. When receiving a datagram
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -141,8 +141,8 @@ for example by having the length in question be a compile-time constant.
 Implementations, which want to use variable-length CIDs, are responsible
 for constructing the CID in such a way that its length can be determined
 on reception. Such implementations must still be able to send
-CIDs of different length to other parties. Is not possible to parse 
-the records without knowing how long the CID is.
+CIDs of different length to other parties. Note that there is no CID 
+length information included in the record itself.
 
 In DTLS 1.2, CIDs are exchanged at the beginning of the DTLS
 session only. There is no dedicated "CID update" message
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -97,8 +97,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
 document are to be interpreted as described in RFC 2119 {{RFC2119}}.
 
-The reader is assumed to be familiar with DTLS {{RFC6347}}.
-
+The reader is assumed to be familiar with DTLS 1.2 {{RFC6347}}.
 
 # The "connection_id" Extension
 
@@ -114,12 +113,12 @@ The extension type is specified as follows.
 ~~~~
 
 The extension_data field of this extension, when included in the
-ClientHello, MUST contain the ConnectionId structure, which carries the CID which
-the client wishes the server to use when sending messages towards it.
-A zero-length value indicates that the client is prepared to send
-with a CID but does not wish the server to use one when
-sending (alternately, this can be interpreted as the client wishes
-the server to use a zero-length CID; the result is the same).
+ClientHello, MUST contain the ConnectionId structure. This structure 
+contains the CID value the client wishes the server to use when sending 
+messages to the client. A zero-length CID value indicates that the client 
+is prepared to send with a CID but does not wish the server to use one when
+sending. Alternatively, this can be interpreted as the client wishes
+the server to use a zero-length CID; the result is the same.
 
 ~~~~
   struct {
@@ -134,34 +133,48 @@ indicates that the server will send with the client's CID but does not
 wish the client to include a CID (or again, alternately, to use a
 zero-length CID).
 
-When a session is resumed, the "connection_id" extension is
-negotiated afresh, not retained from previous connections in
-the session.
-
-Because each party sends the value in the "connection_id" extension that it wants to 
-receive as a connection identifier in encrypted records, it is possible
+Because each party sends the value in the "connection_id" extension it wants to 
+receive as a CID in encrypted records, it is possible
 for an endpoint to use a globally constant length for such connection
 identifiers.  This can in turn ease parsing and connection lookup,
-
 for example by having the length in question be a compile-time constant.
 Implementations, which want to use variable-length CIDs, are responsible
 for constructing the CID in such a way that its length can be determined
-on reception. Note that such implementations must still be able to send
-connection identifiers of different length to other parties.
-
-Note that it is not possible to parse the records without knowing how 
-long the CID is.
+on reception. Such implementations must still be able to send
+CIDs of different length to other parties. Is not possible to parse 
+the records without knowing how long the CID is.
 
 In DTLS 1.2, CIDs are exchanged at the beginning of the DTLS
 session only. There is no dedicated "CID update" message
 that allows new CIDs to be established mid-session, because
 DTLS 1.2 in general does not allow TLS 1.3-style post-handshake messages
-that do not themselves begin other handshakes. 
+that do not themselves begin other handshakes. When a DTLS session is 
+resumed, the "connection_id" extension is negotiated afresh. 
 
-DTLS peers switch to the new record layer format, i.e., the record layer format 
-containing the CID, when encryption is enabled.
+If DTLS peers have negotiated the use of a CIDs using the ClientHello and
+the ServerHello messages then the peers need to take two steps: 
 
-# Record Layer Extensions and Record Payload Protection
+1. First, the peers need to determine whether incoming and ongoing messages need 
+to use the new record layer format, i.e., the record layer format 
+containing the CID. This new record layer format is only used once encryption 
+is enabled. Plaintext payloads never contain a CID in the record layer.
+Whenever a zero-length CID has been negotiated then the RFC 6347-defined 
+record format MUST be used (see Section 4.1 of {{RFC6347}}). For example, 
+in {{dtls-example2}} the use of a CID has been successfully 
+negotiated between a client and a server whereby the client uses 
+the record format defined in this specification to transmit the CID value 
+(100) to the server. The server still uses the RFC 6347-defined 
+record format without a CID when transmitting records to the client. 
+
+2. Second, peers need to use the new MAC calculation defined in this document. 
+Any record layer payload that uses the new record layer format MUST use the 
+new MAC computation defined in {{mac}}. Whenever the RFC 6347-defined 
+record format is used, i.e., no CID is included in the record layer, then 
+the MAC calculation defined in Section 4.1.2 of {{RFC6347}} (and Section 4.1.2.4 
+of {{RFC6347} for use with AEAD ciphers in particular).
+
+
+# Record Layer Extensions
 
 This specification defines the DTLS 1.2 record layer format and 
 {{I-D.ietf-tls-dtls13}} specifies how to carry the CID in DTLS 1.3.
@@ -171,7 +184,7 @@ connections which have negotiated this extension use a distinguished
 record type tls12_cid(25). Use of this content type has the following
 two implications:
 
-- The CID field is present
+- The CID field is present in the record
 - The true content type is inside the encryption envelope, as described
   below.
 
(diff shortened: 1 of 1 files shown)


hannestschofenig:
Completely re-wrote PR #35

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -134,34 +133,48 @@ indicates that the server will send with the client's CID but does not
...
 Implementations, which want to use variable-length CIDs, are responsible
 for constructing the CID in such a way that its length can be determined
-on reception. Note that such implementations must still be able to send
-connection identifiers of different length to other parties.
-
-Note that it is not possible to parse the records without knowing how 
-long the CID is.
+on reception. Such implementations must still be able to send
+CIDs of different length to other parties. Is not possible to parse 
+the records without knowing how long the CID is.
```
boaks: I'm not sure, why "Is not possible to parse the records without knowing how long the CID is." is kept. I tried to address my concerns in #27 . But this may be considered as independent from this PR.

boaks:
Great work! So alternative I made it, right?

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: Maybe it would be better to say that there is no CID length information included in the record itself. I think this would clarify the issue. Do you agree?

boaks: Yes, that sounds good.

hannestschofenig:
We now have to wait for Martin, Thomas, and Ekr to agree that the proposed changes are inline with their thinking and then I would submit a new version of the draft for the working group to look at.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -134,34 +133,48 @@ indicates that the server will send with the client's CID but does not
...
 that allows new CIDs to be established mid-session, because
 DTLS 1.2 in general does not allow TLS 1.3-style post-handshake messages
-that do not themselves begin other handshakes. 
-
-DTLS peers switch to the new record layer format, i.e., the record layer format 
-containing the CID, when encryption is enabled.
-
-# Record Layer Extensions and Record Payload Protection
+that do not themselves begin other handshakes. When a DTLS session is 
+resumed, the "connection_id" extension is negotiated afresh.
```
martinthomson: What about renegotiation?

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -134,34 +133,48 @@ indicates that the server will send with the client's CID but does not
...
+
+If DTLS peers have negotiated the use of a CIDs using the ClientHello and
+the ServerHello messages then the peers need to take two steps: 
+
+1. First, the peers need to determine whether incoming and ongoing messages need 
+to use the new record layer format, i.e., the record layer format 
+containing the CID. This new record layer format is only used once encryption 
+is enabled. Plaintext payloads never contain a CID in the record layer.
+Whenever a zero-length CID has been negotiated then the RFC 6347-defined 
+record format MUST be used (see Section 4.1 of {{RFC6347}}). For example,
```
martinthomson: 6347 record format AND content type

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -134,34 +133,48 @@ indicates that the server will send with the client's CID but does not
...
+Whenever a zero-length CID has been negotiated then the RFC 6347-defined 
+record format MUST be used (see Section 4.1 of {{RFC6347}}). For example, 
+in {{dtls-example2}} the use of a CID has been successfully 
+negotiated between a client and a server whereby the client uses 
+the record format defined in this specification to transmit the CID value 
+(100) to the server. The server still uses the RFC 6347-defined 
+record format without a CID when transmitting records to the client. 
+
+2. Second, peers need to use the new MAC calculation defined in this document. 
+Any record layer payload that uses the new record layer format MUST use the
```
martinthomson: And content type

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -171,7 +184,7 @@ connections which have negotiated this extension use a distinguished
 record type tls12_cid(25). Use of this content type has the following
 two implications:
 
-- The CID field is present
+- The CID field is present in the record
 - The true content type is inside the encryption envelope, as described
   below.
```
martinthomson, line 198: + - The MAC calculation follows the process described in {{}}.

martinthomson (approved):
So I would slightly prefer to have zero- length CID mean use of the new content type, with the associated privacy enhancement, but this is fine.

hannestschofenig:
Martin wrote: 

> I would slightly prefer to have zero- length CID mean use of the new content type, with the associated privacy enhancement

So proposal has not been made so far. We should discuss this idea. This is a new privacy-enhancing feature that does not necessarily have anything to do with the CID idea. In some sense this would allow the client and the server to negotiate the use of a traffic obfuscation technique.

Achim, Ekr, Martin, etc. -- what do you think?

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: Good question. We don't use renegotiation in IoT deployments. It would be interesting to hear from someone who uses renegotation in the IoT context about what the expected behavior should be.

boaks:
> So proposal has not been made so far.

That's the discussions in issues #22 and #43  
I linked to that issue #43 in the previous PR #35 as "Alternative III"
https://github.com/tlswg/dtls-conn-id/pull/35#issuecomment-460030833

To follow up the discussion is really difficult and time intensive. There is also one change left here
https://github.com/tlswg/dtls-conn-id/issues/26#issuecomment-459706334

That request to use the first CID not in FINISH and instead in the record after that.

boaks:
> This is a new privacy-enhancing feature that does not necessarily have anything to do with the CID idea. 

I agree, this adds a new use-case to this draft and is not related to the CID idea.
It's related to the usage of the new introduced record type.

I created a issue about the scope of this draft #40 , which mentions that additional use-case. 
In the meantime I closed that scope issue. My current impression is, that the discussions are split and really hard to follow. Therefore I closed my additional issues in hope, that focus on the CID idea makes it possible to "release" this draft sooner.

hannestschofenig:
You are correct. The proposal has been made before. I was confused....

hannestschofenig:
Achim, do you prefer to use "alternative III" instead of what I wrote-up today?

hannestschofenig:
Btw, I added notes to the draft about the additional privacy features. So, I believe I captured the scope issue you raised.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: Updated text

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: updated text

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: updated text

boaks:
> Achim, do you prefer to use "alternative III" instead of what I wrote-up today?

I prefer any of the solutions! The difference between the flavours is for me much smaller, than the difference in shifting the timelines would be.

> Ekr, Martin, etc. -- 

For me, Hannes did a really great job. 
It would be also great, if you can give your "final" votes to give him a chance to finalize his work :-).

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -376,9 +392,9 @@ Importantly, the sequence number makes it possible for a passive attacker
 to correlate packets across CID changes. Thus, even if a client/server pair
 do a rehandshake to change CID, that does not provide much privacy benefit.
 
-This document does not change the security properties of DTLS {{RFC6347}}.
-It merely provides a more robust mechanism for associating an incoming packet
-with a stored security context.
+The CID-enhanced record layer introduces record padding; a privacy feature 
+not available with the original DTLS 1.2 RFC. Padding allows to inflate the 
+size of the ciphertext making traffic analysis more difficult.
```
boaks: Just to mention:
For that feature it would make sense, to use the new record type even if the CID is empty.
That's what the issues #43 and the discussion in #22 is about.

As already commented, I have no preference.

ekr:
I can live with that proposal as well, as long as we agree that it's a
violation to send an empty CID when one is expected.

boaks:
> I can live with that proposal as well, as long as we agree that it's a violation to send an empty CID when one is expected.

The idea was to use the new record type even if an empty CID is negotiated. 
If a none empty CID is negotiated, then this none empty CID has to be used. 
The mentioned "violation" was, at least from my understand, not intended.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -134,44 +133,59 @@ indicates that the server will send with the client's CID but does not
...
-containing the CID, when encryption is enabled.
-
-# Record Layer Extensions and Record Payload Protection
+that do not themselves begin other handshakes. When a DTLS session is 
+resumed, the "connection_id" extension is negotiated afresh. 
+
+If DTLS peers have negotiated the use of a CIDs using the ClientHello and
+the ServerHello messages then the peers need to take two steps: 
+
+1. First, the peers need to determine whether incoming and ongoing messages need
```
thomas-fossati: s/ongoing/outgoing/

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -134,44 +133,59 @@ indicates that the server will send with the client's CID but does not
...
-# Record Layer Extensions and Record Payload Protection
+that do not themselves begin other handshakes. When a DTLS session is 
+resumed, the "connection_id" extension is negotiated afresh. 
+
+If DTLS peers have negotiated the use of a CIDs using the ClientHello and
+the ServerHello messages then the peers need to take two steps: 
+
+1. First, the peers need to determine whether incoming and ongoing messages need 
+to use the new record layer format, i.e., the record layer format 
+containing the CID. This new record layer format is only used once encryption
```
thomas-fossati: To be sure I'd make this a MUST (and the following a MUST NOT) as this is necessary to interop.

I'd also add some more precision about the exact message where we expect the first CID to appear.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: What about saying for now "When a DTLS session is resumed or renegotiated, the "connection_id" extension is negotiated afresh."?

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: OK. Fixed.

boaks:
Fine! 
AFAIK all discussion about features/scope are concluded with that!
FMPOV #43, #23. #22 are fixed, when this PR is merged.

#18 will then be the only open issue, asking for implementations :-).

boaks:
Uups, I think, I got a very late catch:

https://github.com/tlswg/dtls-conn-id/blob/master/draft-ietf-tls-dtls-connection-id.md#record-layer-extensions-and-record-payload-protection

> cid_length : 1 byte field indicating the length of the negotiated CID. If a zero-length CID has been negotiated, and therefore no CID appears on the wire, a cid_length of zero (0) MUST be added.

**If a zero-length CID has been negotiated, and therefore no CID appears on the wire, a cid_length of zero (0) MUST be added.** 

Is not longer valid!

For a empty CID, the RFC6347 record is used with the RFC6347 MAC and not the new MAC.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -134,44 +133,67 @@ indicates that the server will send with the client's CID but does not
...
 In DTLS 1.2, CIDs are exchanged at the beginning of the DTLS
 session only. There is no dedicated "CID update" message
 that allows new CIDs to be established mid-session, because
 DTLS 1.2 in general does not allow TLS 1.3-style post-handshake messages
-that do not themselves begin other handshakes. 
+that do not themselves begin other handshakes. When a DTLS session is 
+resumed or renegotiated, the "connection_id" extension is negotiated afresh. 
+
+If DTLS peers have not negotiated the use of a CIDs then the RFC 6347-defined 
+record format and content type MUST be used.
```
martinthomson, line 155: Add the zero-length CID thing here.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -134,44 +133,67 @@ indicates that the server will send with the client's CID but does not
...
 that allows new CIDs to be established mid-session, because
 DTLS 1.2 in general does not allow TLS 1.3-style post-handshake messages
-that do not themselves begin other handshakes. 
+that do not themselves begin other handshakes. When a DTLS session is 
+resumed or renegotiated, the "connection_id" extension is negotiated afresh. 
+
+If DTLS peers have not negotiated the use of a CIDs then the RFC 6347-defined 
+record format and content type MUST be used. 
+
+If DTLS peers have negotiated the use of a CIDs using the ClientHello and
```
martinthomson, line 157: Add non-zero-length here.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -134,44 +133,67 @@ indicates that the server will send with the client's CID but does not
...
+the ServerHello messages then the peers need to take the following steps.
 
-DTLS peers switch to the new record layer format, i.e., the record layer format 
-containing the CID, when encryption is enabled.
+The DTLS peers determine whether incoming and outgoing messages need 
+to use the new record format, i.e., the record format containing the CID.
```
martinthomson, line 161: With the above adjustments, this is simply: DTLS peers use the new content type, record format, and MAC calculation for all protected records.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -134,44 +133,67 @@ indicates that the server will send with the client's CID but does not
...
-DTLS peers switch to the new record layer format, i.e., the record layer format 
-containing the CID, when encryption is enabled.
+The DTLS peers determine whether incoming and outgoing messages need 
+to use the new record format, i.e., the record format containing the CID. 
+The new record format and the CID content type is only used once encryption 
+is enabled. Plaintext payloads never use the new record type and the CID content 
+type. 
 
-# Record Layer Extensions and Record Payload Protection
+For sending, if a zero-length CID has been negotiated then the RFC 6347-defined
```
martinthomson, line 166: Why limit this with "for sending"?

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -134,44 +133,67 @@ indicates that the server will send with the client's CID but does not
...
 {{I-D.ietf-tls-dtls13}} specifies how to carry the CID in DTLS 1.3.
 
 In order to allow a receiver to determine whether a record has CID or not,
 connections which have negotiated this extension use a distinguished
 record type tls12_cid(25). Use of this content type has the following
-two implications:
+three implications:
 
-- The CID field is present
+- The CID field is present.
```
martinthomson:
```suggestion
- The CID field is present and contains one or more bytes.
```

draft-ietf-tls-dtls-connection-id.md
boaks: Please explain "all protected records". 
A RFC6347 record may be considered to be protected as well.

martinthomson: Pay special attention to the context.  Which is when a non-zero-length CID is negotiated.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: I wrote the text in a way that it talks about sending and receiving. This is just a writing style.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: I don't think I need to add extra text at this point in the description since I talk about the non-zero length case a few sentences later.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: We could try to shorten the text but currently the style is similar to how Ekr described our conclusion in his email. I liked the way of separating the sending and the receiving processing description.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: But we are not going to add extra functionality in DTLS 1.2 when we already have the same features in 1.3

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -97,8 +97,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
 document are to be interpreted as described in RFC 2119 {{RFC2119}}.
 
-The reader is assumed to be familiar with DTLS {{RFC6347}}.
-
+The reader is assumed to be familiar with DTLS 1.2 {{RFC6347}}.
```
chris-wood: nit: perhaps rewrite to "This document assumes familiarity with DTLS 1.2 {{RFC6347}}."?

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -134,44 +133,67 @@ indicates that the server will send with the client's CID but does not
...
+When transmitting a datagram with the new record format and the new content type, 
+the new MAC computation defined in {{mac}} MUST be used.
+
+For receiving, if the tls12_cid content type is set, then the CID is used to look up 
+the connection and the security association. If the tls12_cid content type is not set, 
+then the connection and security association is looked up by the 5-tuple and a 
+check MUST be made to determine whether the expected CID value is indeed 
+zero length. If the check fails, then the datagram MUST be dropped. 
+
+When receiving a datagram with the new record format and the new content type,
```
chris-wood: nit: s/and the new content type/and tls12_cid content type

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -134,44 +133,67 @@ indicates that the server will send with the client's CID but does not
...
+else the new record layer format and the new content type defined in {{dtls-record12}} MUST be used. 
+
+When transmitting a datagram with the new record format and the new content type, 
+the new MAC computation defined in {{mac}} MUST be used.
+
+For receiving, if the tls12_cid content type is set, then the CID is used to look up 
+the connection and the security association. If the tls12_cid content type is not set, 
+then the connection and security association is looked up by the 5-tuple and a 
+check MUST be made to determine whether the expected CID value is indeed 
+zero length. If the check fails, then the datagram MUST be dropped.
```
chris-wood, line 177: Rather than check for a zero length CID, should we say that the recipient datagram MUST adhere to the format in RFC6347, else be dropped?

draft-ietf-tls-dtls-connection-id.md
chris-wood: When reading this condition one might assume that it's possible to receive a datagram with the new record format but without the new content type. Should we instead fold this together and say that (a) if the new content type is used then the new record format MUST also be in effect, and (b) if the new content type is not used then the old record format MUST be used?

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -376,9 +397,9 @@ Importantly, the sequence number makes it possible for a passive attacker
 to correlate packets across CID changes. Thus, even if a client/server pair
 do a rehandshake to change CID, that does not provide much privacy benefit.
 
-This document does not change the security properties of DTLS {{RFC6347}}.
-It merely provides a more robust mechanism for associating an incoming packet
-with a stored security context.
+The CID-enhanced record layer introduces record padding; a privacy feature 
+not available with the original DTLS 1.2 RFC. Padding allows to inflate the 
+size of the ciphertext making traffic analysis more difficult.
```
chris-wood: Should we reference RFC8446 for more details about this padding and its traffic analysis implications?

chris-wood (changes requested):

draft-ietf-tls-dtls-connection-id.md
boaks: The check is done against the associated "connection and security association" (the negotiated CID), not the received record.

chris-wood: Ah! Thanks for the clarification.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: Yes. A reference to RFC8446 makes sense here.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: Sounds good.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: Fixed.

hannestschofenig: I tried to change the wording to make it more clear that the new record format and the new content type is actually the same. There is no new record format that does not contain the new content type. 

Have a look at the new text

chris-wood (approved):

Issue #46, opened 2019-03-01
CID and non-AEAD ciphersuites


thomas-fossati:
We currently only describe CID processing with AEAD ciphersuites, which seems to imply that the CID extension effectively influences ciphersuite negotiation.

The draft should explain how the two - ciphersuite negotiation and CID ext - interact.

boaks:
I'm not sure, what the scope of this explanation should be?
For the MAC, it seems to be obvious, that the block cipher (https://tools.ietf.org/html/rfc5246#section-6.2.3.2 ) extend the data for the MAC accordingly.

https://tools.ietf.org/html/rfc5246#section-6.2.3.1

```
MAC(MAC_write_key, seq_num +
TLSCompressed.type +
TLSCompressed.version +
TLSCompressed.length +
TLSCompressed.fragment);
```

will get 

```
MAC(MAC_write_key, seq_num +
TLSCompressed.type +
TLSCompressed.version +
TLSCompressed.cid + 
cid_length +
TLSCompressed.length +
TLSCompressed.fragment);
```

That was already discussed in https://github.com/tlswg/dtls-conn-id/pull/29#issuecomment-459277192

> Then, Thomas can make a contribution for the MAC calculation for non-AEAD ciphers.

So, I'm not sure, if you are the intended "Thomas", which may write that :-)

hannestschofenig:
The question is whether we should define the CID-enhanced MAC calculation for non-AEAD ciphers. The current draft only supports AEAD ciphers. If that's OK then we could just tell implementers that the ciphersuite selection should be influenced by this. Alternatively, we could define the MAC calculation also for non-AEAD ciphers. 

What would you prefer, Achim?

ekr:
On Wed, Mar 6, 2019 at 8:26 AM Hannes Tschofenig <notifications@github.com>
wrote:

> The question is whether we should define the CID-enhanced MAC calculation
> for non-AEAD ciphers. The current draft only supports AEAD ciphers. If
> that's OK then we could just tell implementers that the ciphersuite
> selection should be influenced by this. Alternatively, we could define the
> MAC calculation also for non-AEAD ciphers.
>

The latter, I think...

-Ekr

What would you prefer, Achim?

boaks:
Though Leshan (LWM2M, also supporting CBC) is one of the major using and contributing projects, I would also prefer to have the MAC defined for the none-AEAD.

hannestschofenig:
[...]

PR #47, opened 2019-03-07
Added MAC calculation for non-AEAD ciphers

hannestschofenig:
... based on https://github.com/tlswg/dtls-conn-id/issues/46

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -277,9 +277,11 @@ not make use of the DTLSCompressed structure.
 # Record Payload Protection {#mac}
 
 This specification changes the MAC calculation defined in Section 4.1.2 of 
-RFC 6347. At the time of writing ciphers using authenticated 
-encryption with additional data (AEAD) were state-of-the-art. Hence, this 
-specification updates only the additional data calculation defined in 
+RFC 6347. 
+
+## AEAD Ciphers
+
+This  specification updates the additional data calculation defined in 
 Section 6.2.3.3 of {{RFC5246}}, which is re-used by Section
 4.1.2.1 of {{RFC6347}}. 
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -303,14 +305,45 @@ type
...
+
+The MAC is generated as:
+
+   MAC(MAC_write_key, seq_num +
+                        TLSCompressed.type +
+                        TLSCompressed.version +
+						cid +
+						cid_length + 
+                        TLSCompressed.length +
+                        TLSCompressed.fragment);
```
boaks, line 331: My feeling: for the terms used in this draft, `TLSCompressed` should be replaced by `DTLSCiphertext`, and the `cid` is a field of that.

(So, sorry for my irritating comment https://github.com/tlswg/dtls-conn-id/issues/46#issuecomment-468750119. My intention was just to show, that the definitions for the modes are very similar.)

thomas-fossati, line 331: Hi Hannes

given the TLS presentation language typically doesn't add clarity, can I suggest a different approach?  Could we just say that 'cid' and 'cid_length' are inserted between DTLSCiphertext.version and DTLSCiphertext.length in the MAC computations defined by RFC 6347 and RFC 7366 (EtM), and similarly to the authentication data defined in  RFC 6347?

jsalowey: When you agree on an approach please post to WGLC thread on the list.

hannestschofenig: I believe I actually did it wrong. I should have only extended the MAC calculation in RFC 7366.

hannestschofenig: Since we only define this for DTLS 1.2 the calculation should be: 

   MAC(MAC_write_key, seq_num +
       TLSCipherText.type +
       TLSCipherText.version +
       cid + 
       cid_len + 
       TLSCipherText.length +
       IV +
       ENC(content + padding + padding_length));

Correct?

This also means that you have to use the Encrypt-then-MAC extension when using CBC ciphers.

boaks: My concern is still:
This draft uses terms/definitions (`DTLSPlaintext`, `DTLSInnerPlaintext`, and `DTLSCiphertext`) and I would prefer to stay with these terms instead of cite new terms (and details) of other RFC's.

> Could we just say that 'cid' and 'cid_length' are inserted between DTLSCiphertext.version and DTLSCiphertext.length in the MAC computations 

That would be a generic solution, but I'm a little skeptical, which case this would cover, what the approach of this PR, adding the cid to the previous definitions, doesn't.

boaks: @ekr (sorry, I can not resist:)
 
and you're still sure, that this discussions about including the cid into the MAC for 

https://github.com/tlswg/dtls-conn-id/issues/25#issuecomment-444890710

> Yes, but it might be true with some extension in the future. Moreover, authenticating the whole header is just good practice. 

pays off?

ekr: I'm not sure why you think this is a useful question. The burden on spec authors is not really significant compared to the cost of having the spec wrong.

boaks: > I'm not sure why you think this is a useful question.

In #25 I tried to point out, that in my opinion, changing the MAC is not required. The final answer was, it's good practice. So the effort of all the discussions about that is paid for good practice. In my opinion, when good practice gets that expensive, the question, if the good practice really pays off, is a valid question. I was interested in your opinion about that, therefore I asked.

ekr: Opening with "I can not resist" hardly gives the impression that this is a legitimate question rather than trolling. In any case, I don't think this is particularly expensive and my opinion is unchanged.

hannestschofenig:
Closing this PR since Thomas prepared a more comprehensive version in #49

Issue #54, opened 2019-03-12
Privacy


hannestschofenig:
Martin wrote: 
https://mailarchive.ietf.org/arch/msg/tls/yG4msH0W6dlR-92FjwaACyvkT6Y

> I would strike ", if these privacy properties are important in a given deployment" from the acknowledgments section (which is an odd place for the accompanying statement. 

> I would add an explicit note about the lack of CID update making this unsuitable for mobility scenarios.  That's a common use case for this sort of mechanism, but this design lacks any defense against linkability.

PR #56, opened 2019-04-07
Privacy Text Update

hannestschofenig:
Based on https://github.com/tlswg/dtls-conn-id/issues/54

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -400,7 +400,8 @@ communication).  Without multi-homing or mobility, the use of the CID
 is not different to the use of the 5-tuple.
 
 With multi-homing, an adversary is able to correlate the communication
-interaction over the two paths, which adds further privacy concerns.
+interaction over the two paths, which adds further privacy concerns. The lack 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -400,7 +400,8 @@ communication).  Without multi-homing or mobility, the use of the CID
 is not different to the use of the 5-tuple.
 
 With multi-homing, an adversary is able to correlate the communication
-interaction over the two paths, which adds further privacy concerns.
+interaction over the two paths, which adds further privacy concerns. The lack 
+of a CID update mechanism makes this extension unsuitable for mobility scenarios.
```
boaks: I would prefer, to make the "unsuitable" conditional. Therefore I would propose to add something as "when this possible correlation must be considered" . 

My concern here is, that with cid you can communicate perfectly in mobility scenarios, but you may get tracked. Without the condition someone easily misunderstands the "unsuitable", even if the scenario doesn't require "track protection".

boaks:
Thanks!

jsalowey (approved):

Issue #50, opened 2019-03-11
In DTLSPlaintext.version inclusion in MAC intended?


boaks:
https://github.com/tlswg/dtls-conn-id/blob/master/draft-ietf-tls-dtls-connection-id.md#record-payload-protection-mac

```
    Block Ciphers with Encrypt-then-MAC processing:

    MAC(MAC_write_key, seq_num +
        DTLSCipherText.type +
        DTLSCipherText.version +
        DTLSPlaintext.version +
        cid +                   // New input
        cid_length +            // New input
        length of (IV + DTLSCiphertext.enc_content) +
        IV +
        DTLSCiphertext.enc_content);
```

contains two versions, `DTLSCipherText.version` and `DTLSPlaintext.version`.

Is this intended?

boaks:
Fixed with #57

PR #57, opened 2019-04-07
Removed duplicate version from Block Ciphers with ETM MAC calculation

hannestschofenig:
Addresses https://github.com/tlswg/dtls-conn-id/issues/50


boaks:
Thanks! The PR fixes the typo.

jsalowey (approved):

Issue #52, opened 2019-03-11
Where: cid : Value of the negotiated CID.


boaks:
In my opinion, it's the actual cid on the record header, not the negotiated cid, even if both are intended to be the same. Therefore it is `DTLSCiphertext.cid`.

Or has this changed? Is the MAC now extended by the negotiated CID?

boaks:
Fixed with #59

PR #59, opened 2019-04-07
Clarifications of the CID value in the MAC calculation

hannestschofenig:
Addresses https://github.com/tlswg/dtls-conn-id/issues/52


boaks:
Thanks! This PR clarifies the used cid value.

jsalowey (approved):

PR #61, opened 2019-04-09
Left connection_id and tls12_cid for IANA to assign

hannestschofenig:


boaks:
https://mailarchive.ietf.org/arch/msg/tls/TTXYlad2SExikFKwV2EuaEKvf1Q

If this PR is postponed to that date (12. April), may be the final values could be used.

jsalowey (approved):

Issue #64, opened 2019-05-24
Peer address validation


thomas-fossati:
Alice has an active DTLS session + CID with Bob.
  - Eve (man-on-the-side) sends junk packet with the CID currently in use between Alice and Bob and a spoofed source address of Victor
      - If Bob updates the peer address before the record is validated, he creates DoS/DDoS surface.
  - Eve (man-on-the-side) replays previous packet from Alice and changes source address to that of Victor
      - If Bob has anti-replay disabled — even if he validates the records in the packet — he creates DoS/DDoS surface.
  - Eve (man-in-the-middle) modifies packet from Alice to have the source address of Victor
    - Expanded DoS/DDoS surface is unavoidable.

While the first two can be addressed, I don't know how to stop (or even mitigate the third).

We should add text to the SecCons along the following lines:
```
[...]
When using CID, anti-replay MUST be enabled.
When a record with CID is received, the receiver MUST NOT update
its view of the peer address with the source specified in the UDP packet
before cryptographically validating the enclosed record(s).  Failing to do
so, creates a (D)DoS opportunity for a man-on-the-side attacker.
Use of CID creates the opportunity for an active man-in-the-middle to
trigger backscatter traffic for each CID-bearing packet it can modify.
[hic sunt leones]
```

thomas-fossati:
We have the same problem in 1.3, but there the active man-in-the-middle attack described above could be mitigated by triggering an "address validation" sub protocol (via post-handshake) either on each and every peer address change or when the rate of peer address changes is above a certain threshold.

boaks:
I would assume, that such an attack makes only sense, if an "amplified response" is expected. Is that the right assumption?
That seems to be out of the scope of DTLS (means DTLS doesn't define such a response).
Therefore I think, it could only be recommended, that a DTLS CID implementation tracks, if the associated ip-address should be considered as "verified". Each address change would reset that, and only a ["Heartbeat"](https://tools.ietf.org/html/rfc6520) (if supported), or the next layer could set it as "verified" again. The DTLS CID implementation then limits the maximum application message size for unverified destinations and reports a violation as error. It depends on the "next layers", if and how such a verification will be done (may be by explicit trigger a "heartbeat", if supported).
Using CoAP, I would assume, that a lot of outgoing traffic is just a ACK/SUCCESS, which is pretty small and so requires no special behaviour. For larger outgoing messages [echo-request-tag](
[...]

PR #65, opened 2019-06-01
peer address update checks

thomas-fossati:
A first attempt to address #64 , as per discussion with Achim.

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -347,27 +347,41 @@ data the following modification is made to the additional data calculation.
 
 # Peer Address Update Checks
 
+A priori, a receiver has no means to distinguish an address update event that
+is genuine (for example, resulting from a NAT rebinding occurring at an on-path
+middlebox) from one that is malicious (for example, a man-on-the-side /
+man-in-the-middle attacker trying to either DoS the receiver’s peer or use the
+receiver as a DDoS reflector.)
+
+This section describes what an implementation must do in order to thwart, or at
+least mitigate, an attacker’s attempts at maliciously subverting the
+functionality provided by CID by manipulating the source 2-tuple of CID bearing
+datagrams.
+
 When a record with CID is received that has the source address of the enclosing
-UDP datagram different from the one previously recorded, the receiver MUST NOT
-update its view of the peer address with the source specified in the UDP packet
-before cryptographically validating the enclosed record(s).
+UDP datagram different from the one previously associated with that CID, the
+receiver MUST NOT update its view of the peer’s address with the source
+specified in the UDP packet before cryptographically validating the enclosed
+record(s).  This is to ensure that a man-on-the-side attacker that sends a
+packet with a different source address on an existing CID session does not
+successfully manage to reroute any return traffic.
 
-Immediately after reception, the recipient SHOULD initiate address validation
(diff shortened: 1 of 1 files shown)


ekr:
I think the first paragraph of this PR is fine, but the rest is out of scope.

The agreed upon intent of this draft is simply to allow flow identification, not to provide a generic address migration facility. That is a complicated topic that should be out of scope here.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -345,6 +345,30 @@ data the following modification is made to the additional data calculation.
...
+
+When a record with CID is received that has the source address of the enclosing
+UDP datagram different from the one previously recorded, the receiver MUST NOT
+update its view of the peer address with the source specified in the UDP packet
+before cryptographically validating the enclosed record(s).
+
+Immediately after reception, the recipient SHOULD initiate address validation
+using a Heartbeat {!RFC6520} based ping-pong with the new peer address.  This
+is to ensure that address validation can be cryptographically verified.  Both
+Heartbeat messages (request and response) MUST be wrapped into tls12_cid
```
boaks: FMPOV, **MUST** for `request` is (currently) not possible, because the use of tls12_cid is bound to a none empty cid. Only the response can considered to be MUST tls12_cid, because the cause of this address-issues is bound to use such a none empty-cid for incoming messages.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -345,6 +345,30 @@ data the following modification is made to the additional data calculation.
...
+records.  If CID has not been negotiated, the sender MUST use a 0-length CID
+instead of the RFC 6347-defined record format. While the address validation
```
boaks: My feeling is, tls12_cid with empty cid was discussed some months ago. To change this, would cause changes in other parts (MAC), and I'm not sure, if this pays off. Assuming, that the response will be a tls12_cid, which risk do you see, which cause the MUST for the request?

boaks:
I would prefer / appreciate, if the parts after the first paragraph would start with explaining the vulnerability and the attack, which may use it. That would setup the right scope for understanding, when a mitigation is required to protect other systems.

In the discussion #64 I tried to point out, that I would prefer to leave the mitigation to the using system and just mention the "heartbeat" as one solution.
As I wrote in a comment above, I'm not sure, if introducing "tls12_cid for empty cid" will pay off. May be you can provide us the scenario, where the request requires tls12_cid as well.

jsalowey:
We don't have consensus for all of this text.  I think the proposal is to address tis in a separate document.  Can we close this PR?

thomas-fossati:
> We don't have consensus for all of this text. I think the proposal is to address tis in a separate document. Can we close this PR?

The current version of this PR has expunged the text that outlined an address validation sub-protocol.  This can and will be sketched separately.  What is left here is just a description of what the problem is and what a receiver must do to counter at least two possible attacks using simple unilateral measures.  I'm not sure what is controversial?

boaks:
I would prefer, if the risk of the "man-in-the-middle" explicitly states, that

- it's a risk to be misused to attack other systems
  (the own system may attacked by such a  "man-in-the-middle" much harder with DoS by just dropping the messages)
- therefore only applies to large response messages

In my opinion, too general descriptions of risks tend to be misunderstood.

hannestschofenig:
This PR has been replaced by text offered in https://github.com/tlswg/dtls-conn-id/pull/68

PR #67, opened 2019-06-17
adding IANA values

### LINKED ISSUE #66: IANA Considerations Update
A couple of tweaks are needed for the IANA Considerations Section:

- Based on input from the IANA TLS DEs, it seems like we need a column added to the TLS ExtensionType Values registry to indicate that an extension is DTLS-only.
- The column values for TLS 1.3 and Recommended are missing.  Suggest blank for TLS 1.3 and N for Recommended.
- Add a note to say that connection_id is DTLS1.2 only.
- Add instructions for IANA to assign "Y" to DTLS-okay column for tls12_cid content type.

seanturner:
We need to tell IANA what to assign to all of the columns.

Closes #66

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -441,10 +441,15 @@ about the padding can be found in Section 5.4 and Appendix E.3 of RFC 8446.
 
 IANA is requested to allocate an entry to the existing TLS "ExtensionType
 Values" registry, defined in {{RFC5246}}, for connection_id(TBD1) defined in
-this document.
+this document.  The "TLS 1.3" column is "-" because the extension is not
+applicable to TLS 1.3 and the "Recommended" column is "Y".
+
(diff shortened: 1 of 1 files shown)


jsalowey:
Looks good

richsalz:
ship it

boaks:
Sorry, I wanted to add this to the issue #66 , not the PR.

++++++++++++++++++++

That results in 

"DTLS only" for TLS extensions and
"DTLS-OK" for TLS parameters.

I'm not sure, if that just represents more the "current" usage than the "logic" behind it. 
I would be afraid, that this will hold up other ideas, which doesn't match that "only for extensions / ok for parameters" pattern.

seanturner:
See:
https://mailarchive.ietf.org/arch/msg/tls/JdKKK2kKumrt4Z-51Fqx-JI9QsU

hannestschofenig:
This PR has been replaced by text offered in https://github.com/tlswg/dtls-conn-id/pull/68

PR #68, opened 2019-07-07
IANA and Security Consideration Sections Updated

hannestschofenig:
... IANA consideration section based on issue raised by Sean (see https://github.com/tlswg/dtls-conn-id/issues/66) 

... Security consideration section based on issue raised by Thomas (see https://github.com/tlswg/dtls-conn-id/issues/64)


hannestschofenig:
This PR aims to replace PR#65 & PR#67

Issue #69, opened 2019-07-10
Preconditions for accepting address update events


boaks:
[Security and Privacy Considerations](https://github.com/tlswg/dtls-conn-id/blob/master/draft-ietf-tls-dtls-connection-id.md#security-and-privacy-considerations-sec-cons)

The part 

> When a record with CID is received that has the source address of the enclosing
> UDP datagram different from the one previously associated with that CID, the
> receiver MUST NOT update its view of the peer’s address with the source
> specified in the UDP packet before cryptographically validating the enclosed
> record(s).  This is to ensure that a man-on-the-side attacker that sends a
> packet with a different source address on an existing CID session does not
> successfully manage to reroute any return traffic.

of PR #65 didn't made to the draft, or did I miss it?

In the meantime, I would even go further:
Applying [MAC](https://tools.ietf.org/html/rfc6347#section-4.1.2.1) and [Anti-Replay](https://tools.ietf.org/html/rfc6347#section-4.1.2.6) may validates the record cryptographically, and so the receive windows may be updated and the application data may be forwarded. But, if the epoch/sequence_number is not  newer than the newest successfully validate epoch/sequence_number before, the address update event itself may be wrong! Assuming, that a address-change is located somewhere on the message's route, it seems to be not possible to distinguish, if the message was delayed after the "address change" (and so that address is deprecated) or before that (and so the address would be the new one). If that could not be decided, I would go for updating the address only for newer records.
Using that approach, updating the address only for newer records (based on epoch/sequence_number) with validated [MAC](https://tools.ietf.org/html/rfc6347#section-4.1.2.1), would also mitigate the first two attacks of issue #64, though retransmitting an record with changed address will not update the address, because these records are **NOT** newer.

hannestschofenig:
I have updated the text of https://github.com/tlswg/dtls-conn-id/pull/70 and was wondering whether it captures your issue. I am not sure what steps I have to take to meet your requirements.

boaks:
As long as [RFC6347, 4.1.2.6 - anti replay](https://tools.ietf.org/html/rfc6347#section-4.1.2.6) doesn't get mandatory (MUST), everything for me will be OK.

So, yes, PR #70 addresses it in the right way.

thomas-fossati:
It looks like #70 closes this issue?

boaks:
Thanks!

PR #70, opened 2019-07-23
Address validation take 2

thomas-fossati:


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -345,6 +345,39 @@ data the following modification is made to the additional data calculation.
                       length_of_DTLSInnerPlaintext;
 ~~~
 
+# Peer Address Update
+
+When a CID record is received with a different source address than the
```
hannestschofenig: How about 

"When a datagram with a CID is received that has a source address different than the one currently associated with the DTLS connection, the receiver MUST NOT replace the source address currently stored in the association with the source address specified in the received datagram unless the following conditions are met:
 * The received datagram has been cryptographically verified using the DTLS record layer processing procedures. 
- The received datagram are "newer" (in terms of their epoch and sequence
  number) than the last datagram that successfully updated the source address.

thomas-fossati: I like it, but I think using the term datagram instead of record when talking about crypto operations mixes layers.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -345,6 +345,33 @@ data the following modification is made to the additional data calculation.
                       length_of_DTLSInnerPlaintext;
 ~~~
 
+# Peer Address Update {#peer-address-update}
+
+When a datagram with a CID is received that has a source address different than the one currently associated with the DTLS connection, the receiver MUST NOT replace the source address currently stored in the association with the source address specified in the received datagram unless the following conditions are met:
```
martinthomson: Rewrapping this only.
```suggestion
When a datagram with a CID is received that has a source address different than
the one currently associated with the DTLS connection, the receiver MUST NOT
replace the source address currently stored in the association with the source
address specified in the received datagram unless the following conditions are
met:
```

However, I would also suggest that when you talk about "source address" (other than the first instance), you instead use "peer address", perhaps as follows:

```suggestion
When a datagram with a CID is received that has a source address different than
the one currently associated with the DTLS connection, the receiver MUST NOT
replace the address it uses for sending datagrams to its peer with the source
address specified in the received datagram unless the following conditions are
met:
```

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -345,6 +345,33 @@ data the following modification is made to the additional data calculation.
...
+
+- The received datagram has been cryptographically verified using the DTLS record layer processing procedures.
```
martinthomson: ...and any duplicate detection (?)

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -345,6 +345,33 @@ data the following modification is made to the additional data calculation.
...
+
+- The received datagram is "newer" (in terms of their epoch and sequence
```
martinthomson:
```suggestion
- The received datagram is "newer" (in terms of both epoch and sequence
```

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -345,6 +345,33 @@ data the following modification is made to the additional data calculation.
...
+number) than the last datagram that successfully updated the source address.
```
martinthomson:
```suggestion
number) than the newest datagram received.
```

for the parenthetical, I would instead use "The received datagram is "newer" than the any other datagram that has been received (meaning that the newer datagram has a higher epoch or a matching epoch and a higher sequence number)"

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -345,38 +345,32 @@ data the following modification is made to the additional data calculation.
...
-
-Since this document does not define an in-protocol peer validation
-procedure, implementations that do not already offer the mechanism
-described in {{!I-D.tschofenig-tls-dtls-rrc}} MUST, on CID-enabled
-sessions, report any peer address update to their users.  When delivered
-such an event, the user can trigger an application layer-specific
-address validation mechanism, for example one that is based on
-successful exchange of minimal amount of ping-pong traffic with the
-peer.
+Implementations that offer an application layer return routability check
```
thomas-fossati: turned this way sounds a bit funny.  I'd rather say that "[...] if the implementation offers *peer address update detection*, it MUST report any peer address update on CID-enabled sessions to applications"

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -345,6 +345,33 @@ data the following modification is made to the additional data calculation.
...
+# Peer Address Update {#peer-address-update}
+
+When a datagram with a CID is received that has a source address different than the one currently associated with the DTLS connection, the receiver MUST NOT replace the source address currently stored in the association with the source address specified in the received datagram unless the following conditions are met:
+
+- The received datagram has been cryptographically verified using the DTLS record layer processing procedures.
+
+- The received datagram is "newer" (in terms of their epoch and sequence
+number) than the last datagram that successfully updated the source address.
+This condition ensures that replayed datagrams cannot be used to perform address 
+updates. Note that this condition does not help when the attacker is able to
```
martinthomson: I think that the main thing here is to observe that reordered datagrams don't cause excessive thrashing if datagrams are reordered by the network.  That makes this less than a firm requirement and maybe not something that you want to include in the list, but more as supplementary information.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -345,6 +345,33 @@ data the following modification is made to the additional data calculation.
...
+
+- The received datagram has been cryptographically verified using the DTLS record layer processing procedures.
+
+- The received datagram is "newer" (in terms of their epoch and sequence
+number) than the last datagram that successfully updated the source address.
+This condition ensures that replayed datagrams cannot be used to perform address 
+updates. Note that this condition does not help when the attacker is able to 
+replay datagrams that arrive earlier at the DTLS peer than the original datagram.
+
+- There is a strategy for performing a perform some reachability test.
```
martinthomson:
```suggestion
- There is a strategy for ensuring that the new peer address is able to receive and process
DTLS datagrams for the connection.
```

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -345,6 +345,33 @@ data the following modification is made to the additional data calculation.
...
+number) than the last datagram that successfully updated the source address.
+This condition ensures that replayed datagrams cannot be used to perform address 
+updates. Note that this condition does not help when the attacker is able to 
+replay datagrams that arrive earlier at the DTLS peer than the original datagram.
+
+- There is a strategy for performing a perform some reachability test. 
+No such test is defined in this	specification; {{!I-D.tschofenig-tls-dtls-rrc}} 
+defines an extension for such a DTLS protocol extension.
+
+The above ensures correctness of the protocol in presence of packet
```
martinthomson: There is no "ensures" here.  And, as mentioned, I think that the reordering thing is an optimization only, the reordering point can be separated.

"The above is necessary to protect against attacks that use datagrams with spoofed addresses or replayed datagrams to trigger attacks."

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -345,6 +345,33 @@ data the following modification is made to the additional data calculation.
...
+reordering at the network layer, while also thwarting a man-on-the-side
+attacker attempting to use spoofed or replayed records to reroute return
+traffic.
+
+Implementations that offer an application layer return routability check
```
martinthomson:
```suggestion
Application protocols that implement protection against these attacks depend on
being aware of changes in peer addresses so that they can engage the necessary
mechanisms.
```

Then remove the remaining lines of the paragraph.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -421,13 +448,16 @@ An on-path adversary, who is able to observe the DTLS protocol exchanges between
 DTLS client and the DTLS server, is able to link the observed payloads to all
 subsequent payloads carrying the same connection id pair (for bi-directional
 communication).  Without multi-homing or mobility, the use of the CID
-is not different to the use of the 5-tuple.
+has the same privacy properties of the 5-tuple.
```
martinthomson:
```suggestion
exposes the same information as the 5-tuple.
```

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -421,13 +448,16 @@ An on-path adversary, who is able to observe the DTLS protocol exchanges between
 DTLS client and the DTLS server, is able to link the observed payloads to all
 subsequent payloads carrying the same connection id pair (for bi-directional
```
martinthomson:
```suggestion
subsequent payloads carrying the same CID pair (for bi-directional
```

martinthomson (commented):
Lots of suggestions.  I hope they are helpful.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: The proposed change looks good to me

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: Good catch on the CID/connection id terminology

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: The proposed changes are good for me

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: Ok for me.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: I am failing to see the changes in the proposed text. I understand why you want to use the term 'peer address' in context and I proposed some text

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: We list the different requirements separately.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: Sounds good to me!

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: Fine for me. I think both text segments are pretty clear about their meaning but I am happy to take this text proposal on board.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: I am not sure what text changes would be useful here because I don't understand the "thrashing aspect" you talk about here.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: Good text for me. I included it.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: I merged your text with the text from Martin and replaced the current paragraph

draft-ietf-tls-dtls-connection-id.md
martinthomson: "Reordered datagrams that are sent prior to a change in a peer address might otherwise cause a valid address change to be reverted. This also limits the ability of an attacker to use replayed datagrams to force a spurious address change, which could result in denial of service, however an attacker might be able to force an address change if they are able to rewrite source addresses or if replayed packets are able to arrive before any original."

?

martinthomson (commented):
The changes look good to me.

draft-ietf-tls-dtls-connection-id.md
boaks: The main trigger for my comment in issue #69 was the previous definition based on "record sequence number window", which then would turn into "MUST use such a window" in order to prevent such simple kind of attacks.

I therefore wrote that comment, to show, that both solutions may have their justification. 

Though you never know, if your under "fire" or if it*s a delay before or after a NAT. Therfore you will never know, what will be the perfect action, to update or not update the address. Neither with the window nor with the newer definition.
My personal preference is therefore the simpler implementation.
An my intention of issue #69 is to prevent this extension from adding a "MUST" to the window approach.

boaks: > This also limits the ability of an attacker to use replayed datagrams

FMPOV, both solutions, the "window" or "newer" will prevent address updates from replay attacks. In difference to a replay attack, where the record sequence number is used twice (and assume, that the modified datagram is received later), such attack are mitigated by both approaches.

boaks: > however an attacker might be able to force an address change if they are able to rewrite source addresses

FMPOV, that's mainly the "on-path-adversary", which requires other means, which are out of the scope of this document.

> or if replayed packets are able to arrive before any original. 

FMPOV, the means against such "lucky attackers" are the same as for the "on-path-adversary".

In my comment in the tls-mailing list 

https://mailarchive.ietf.org/arch/msg/tls/8EAYDGU5RqVcH6POHFZg30wcUYE
  
I expressed, 

> I would prefer to have the basic advice, to apply some kind of filter against modified replay attacks (either the 
> DTLS 1.2 Anti-Replay protection, https://tools.ietf.org/html/rfc6347#section-4.1.2.6, or other means, see
> proposal https://github.com/tlswg/dtls-conn-id/issues/69 ) in the "draft-ietf-tls-dtls-connection-id", 
> instead of including that to an additional RFC together with the more complex "on-path adversary" scenario.


So again, my preference is to have the simple "filter" definitions "window or newer" both mentioned in this document. Also the comment, that "on-path-adversary" will able to bypass that protection (as "lucky" attacker will also do). 
And the discussion and solution for the "advanced on-path-adversary" in a separate document.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -414,20 +440,23 @@ Every identifier introduces the risk of linkability, as explained in {{RFC6973}}
...
 
 An on-path adversary, who is able to observe the DTLS protocol exchanges between the
 DTLS client and the DTLS server, is able to link the observed payloads to all
-subsequent payloads carrying the same connection id pair (for bi-directional
+subsequent payloads carrying the same ID pair (for bi-directional
 communication).  Without multi-homing or mobility, the use of the CID
-is not different to the use of the 5-tuple.
+exposes the same information as the 5-tuple.
 
 An on-path adversary can also black-hole traffic or create a reflection attack
```
boaks: At the risk of repeating it too often:

I would prefer, not to relate the "black-hole" attack of an "on-path-adversary" with the CID!
Until now, nobody presented any reason, why such a general "black-hole" is related to this CID.

So:
- either generalize the threat

"An on-path adversary can not only black-hole traffic, which is generally also possible without the CID, additionally a reflection attack could be created ..."
 
- or remove the "black-hole".

In my experience, such threats, which are for experts obviously not related to a specific technique, will be cited by others, and then it will turn into time consuming discussion instead of spending the time in protection against real threats.

THEREFORE:
Please, don't overstate the threat! The left real threat one is hard enough.

hannestschofenig: Fair enough. That threat is not new to this document and hasn't been a concern to anyone as far as I know. I will remove it.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: I completely agree with you. We don't want to make this artificially complicated.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: Changed the text and removed the term black-hole.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -405,43 +444,51 @@ record layer payload containing the Finished messagen contains a CID.
...
+exposes the same information as the 5-tuple.
+
+With multi-homing, a passive attacker is able to correlate the communication
+interaction over the two paths and the sequence number makes it possible 
+to correlate packets across CID changes. The lack of a CID update mechanism 
+in DTLS 1.2 makes this extension unsuitable for mobility scenarios where 
+correlation must be considered. Deployments that use DTLS in multi-homing
+environments and are concerned about this aspects SHOULD refuse to use CIDs in 
+DTLS 1.2 and switch to DTLS 1.3 where a CID update mechanism is provided and 
+sequence number encryption is offered.
```
chris-wood:
```suggestion
sequence number encryption is available. 
```

chris-wood (approved):
This split makes sense to me. Thanks for making the change!

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -471,6 +518,12 @@ Registry". The tls12_cid ContentType is only applicable to DTLS 1.2.
 
 RFC EDITOR: PLEASE REMOVE THE THIS SECTION
 
+draft-ietf-tls-dtls-connection-id-06
```
boaks: My feeling is, that should be draft-ietf-tls-dtls-connection-id-**07**, or?

PR #71, opened 2020-02-05
Corrected reference to the RRC draft

hannestschofenig:
We incorrectly put a normative reference on the RRC draft in the DTLS 1.2 CID draft.

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -1,7 +1,7 @@
 ---
 title: Connection Identifiers for DTLS 1.2
 abbrev: DTLS 1.2 Connection ID
-docname: draft-ietf-tls-dtls-connection-id-07
+docname: draft-ietf-tls-dtls-connection-id-latest
 category: std
 updates: 6347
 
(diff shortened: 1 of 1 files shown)


boaks:
I lost somehow the track on that RRC and the discussion around it.

The referred document mentions:

> An on-path adversary could intercept and modify the source IP address (and the source port).

and

> Without such a return routability check, an adversary can redirect  traffic towards a third party or a black hole.

I would assume, that a on-path adversary could black hole traffic in both directions, even without CID. Not only message sent to a peer could be dropped, it would also be easy to drop a message sent back from that peer (at least, if the route back is on the same path). 

I would also assume, that a on-path adversary, which is able to change the source address, would also be able to change the destination address. So a simple traffic redirect will also be possible even without CID.

In my opinion, the only threat left, which is caused by the usage of a CID, would be a DDoS attack using amplified traffic towards third party. For that it seems to be most important to throttle the traffic by limiting the number and size of messages until the address is verified. Though such a DDoS will be naturally less efficient, because the adversary has to wait for messages to change the sources-address, throttling would make it even more unattractive.

PR #73, opened 2020-05-13
Disallow sending MAC failure fatal alerts to non-validated peers

### LINKED ISSUE #72: Disallow sending MAC failure fatal alerts to non-validated peers
Imagine the following:
* server S has an active session with client A using CID=1
* S crashes and restarts
* a new session using CID=1 is successfully established between S and client B
* A sends a record with CID=1
* MAC verification fails for the record

If at this point S decides *not* to follow the `SHOULD silently discard` recommendation in [RFC 6347](https://tools.ietf.org/html/rfc6347#section-4.1.2.1) and instead replIes to B (since A's address has not been validated) with a fatal MAC failure alert, a perfectly valid session would be killed.

thomas-fossati:
Fixes #72

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -366,7 +366,10 @@ in changing a peer address if they are able to rewrite source addresses
 and if replayed packets are able to arrive before any original. 
 
 - There is a strategy for ensuring that the new peer address is able to 
-receive and process DTLS records. No such test is defined in this	specification. 
+receive and process DTLS records. No such test is defined in this specification.
+
+If any of the above checks fails, the receiver MUST silently discard the
+record.
 
 The above is necessary to protect against attacks that use datagrams with 
 spoofed addresses or replayed datagrams to trigger attacks. Note that there 
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -352,7 +352,7 @@ different than the one currently associated with the DTLS connection,
 the receiver MUST NOT 
 * replace the address it uses for sending records to its peer with the 
   source address specified in the received datagram, or
-* send a response of any kind to the record, including sending FatalAlerts
+* send a response of any kind to the record, including sending alerts
   in response to an invalid record,
 unless the following conditions are met:
 
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -371,10 +371,7 @@ and if replayed packets are able to arrive before any original.
 - There is a strategy for ensuring that the new peer address is able to 
 receive and process DTLS records. No such test is defined in this specification.
 
-If any of the above conditions fail, the receiver MUST silently discard the
-received datagram.
-
-The above is necessary to protect against attacks that use datagrams with 
+The conditions above are necessary to protect against attacks that use datagrams with 
 spoofed addresses or replayed datagrams to trigger attacks. Note that there 
 is no requirement to use of the anti-replay window mechanism defined in 
 Section 4.1.2.6 of DTLS 1.2. Both solutions, the "anti-replay window" or 
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -349,12 +349,9 @@ data the following modification is made to the additional data calculation.
 
 When a record with a CID is received that has a source address 
 different than the one currently associated with the DTLS connection,
-the receiver MUST NOT 
-* replace the address it uses for sending records to its peer with the 
-  source address specified in the received datagram, or
-* send a response of any kind to the record, including sending alerts
-  in response to an invalid record,
-unless the following conditions are met:
+the receiver MUST NOT replace the address it uses for sending records 
+to its peer with the source address specified in the received datagram
+unless the following three conditions are met:
 
 - The received datagram has been cryptographically verified using 
 the DTLS record layer processing procedures.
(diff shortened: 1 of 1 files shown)


hanno-becker (approved):
Looks good to me

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -366,7 +366,10 @@ in changing a peer address if they are able to rewrite source addresses
 and if replayed packets are able to arrive before any original. 
 
 - There is a strategy for ensuring that the new peer address is able to 
-receive and process DTLS records. No such test is defined in this	specification. 
+receive and process DTLS records. No such test is defined in this specification.
+
+If any of the above checks fails, the receiver MUST silently discard the
```
boaks: This doesn't mention the main fear, that a message is sent to the previous not updated address.

The "any of the above checks" may cause irritation, though the check for the newer address does not inhibit the proper processing of a "out of order" record on application layer.

hanno-becker:
Here's an alternative suggestion - WDYT, @boaks @thomas-fossati. Instead of

```
When a record with a CID is received that has a source address 
different than the one currently associated with the DTLS connection, 	
the receiver MUST NOT replace the address it uses for sending records 	
to its peer with the source address specified in the received to its peer
with the source address specified in the received datagram unless the 
following conditions are met:
```

write:

```
When a record with a CID is received that has a source address 
different than the one currently associated with the DTLS connection,
the receiver MUST NOT 
* replace the address it uses for sending records to its peer with the 
  source address specified in the received datagram, or
* send a response of any kind to the record, including sending FatalAlerts
  in response to an invalid record,
unless the following conditions are met:
```

It's a bit long, though.

thomas-fossati:
> Here's an alternative suggestion - WDYT, @boaks @thomas-fossati.

I don't want to speak for @boaks :-), but I think he's arguing for a wider scope than just _"When a record with a CID is received"_

boaks:
> The received datagram has been cryptographically verified using the DTLS record layer processing procedures. The option of RFC6347, 4.1.2.1, to choose to generate an alert when a message with an invalid MAC is received, MUST NOT be used. The probability of accidentally terminating a valid connection state is increased using this CID. Additionally the destination address to send such an alert is ambiguous, though the address MUST NOT be updated. 
 
FMPOV that describes the increment introducing CID in the scope of MAC validation and alert generation.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -348,10 +348,13 @@ data the following modification is made to the additional data calculation.
...
-the receiver MUST NOT replace the address it uses for sending records 
-to its peer with the source address specified in the received 
-datagram unless the following conditions are met: 
+different than the one currently associated with the DTLS connection,
+the receiver MUST NOT 
+* replace the address it uses for sending records to its peer with the 
+  source address specified in the received datagram, or
+* send a response of any kind to the record, including sending FatalAlerts
+  in response to an invalid record,
+unless the following conditions are met:
```
```suggestion
* send a response of any kind to the record, including sending alerts
  in response to an invalid record,
unless the following conditions are met:
```

draft-ietf-tls-dtls-connection-id.md
ekr: I am also not sure what this sentence means. Whcih checks?

ekr (changes requested):

boaks:
Let me try to point to the difference of the current text and my proposal above:

If the address hasn't changed since the last "validation", should a peer chose the option of [RFC 6347 - MAC](https://tools.ietf.org/html/rfc6347#section-4.1.2.1) to 

> generate a bad_record_mac alert with level fatal and terminate its connection state.

?

If the address has changed, then a peer may choose that option but must not send bad_record_mac alert?

In my opinion, this option to "alert" should not be chosen for CID. Therefore, I would point to that option in RFC 6347 and recommend not to use it with CID.

draft-ietf-tls-dtls-connection-id.md
thomas-fossati: > I am also not sure what this sentence means. Whcih checks?

the three conditions in L357-372

hannestschofenig: I have updated the text to clarify that we are talking about the conditions above.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -366,7 +369,10 @@ in changing a peer address if they are able to rewrite source addresses
 and if replayed packets are able to arrive before any original. 
 
 - There is a strategy for ensuring that the new peer address is able to 
-receive and process DTLS records. No such test is defined in this	specification. 
+receive and process DTLS records. No such test is defined in this specification.
+
+If any of the above conditions fail, the receiver MUST silently discard the
+received datagram.
```
ekr: OK, so there are two different statements here:

1. Don't update the IP and send
2. Don't even accept the packet.

If we want (2) then we can simplify, right?

boaks: Therefore make general case of RFC6347

> In general, DTLS implementations SHOULD silently discard records with bad MACs or that are otherwise invalid.

mandatory for CID and remove the option to alert.

hannestschofenig: I removed the sentence being discussed here.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -366,9 +369,9 @@ in changing a peer address if they are able to rewrite source addresses
 and if replayed packets are able to arrive before any original. 
 
 - There is a strategy for ensuring that the new peer address is able to 
-receive and process DTLS records. No such test is defined in this	specification. 
+receive and process DTLS records. No such test is defined in this specification.
 
-The above is necessary to protect against attacks that use datagrams with 
+The conditions above are necessary to protect against attacks that use datagrams with
```
ekr, line 371: I'm still confused about this. There are two values here:

- Don't change the path
- Don't process the packet

This text, however, seems to allow you to process it but not to respond, even to the original IP. What's the reason for that?

hannestschofenig: I now see what creates the confusion. I have reworded the text and hope it is better now...

ekr: LGTM

ekr (approved):

PR #75, opened 2020-09-03
Editorial bits from AD review

kaduk:
This is supposed to all be editorial stuff, and I split a couple changes out into separate commits in case they're not actually desired.

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -497,8 +497,11 @@ that is malicious. This attack is of concern when there is a large asymmetry
 of request/response message sizes. 
 
 Additionally, an attacker able to observe the data traffic exchanged between 
-two DTLS peers is able to replay datagrams with modified IP address/port numbers
-if the optional DTLS replay detection functionality is not in use. 
+two DTLS peers is able to modify IP address/port numbers without causing the
+packet to be rejected by the recipient.  When the optional DTLS replay
+detection is not in use, such an attacker can also replay the observed packets,
+and the multiple copies of each packet can have different IP address/port
+numbers as well.
 
 The topic of peer address updates is discussed in {{peer-address-update}}.
 
@@ -519,8 +522,8 @@ TBD1    connection_id   -        Y          N           [[This doc]]
 
 Note: The value "N" in the Recommended column is set because this 
 extension is intended only for specific use cases. This document describes 
-an extension for DTLS 1.2 only; it is not applicable to TLS, and
-analogous functionality for DTLS 1.3 is described in {{I-D.ietf-tls-dtls13}}. 
+the behavior of this extension for DTLS 1.2 only; it is not applicable to TLS, and
+its usage for DTLS 1.3 is described in {{I-D.ietf-tls-dtls13}}.
 
 IANA is requested to allocate tls12_cid(TBD2) in the "TLS ContentType
 Registry". The tls12_cid ContentType is only applicable to DTLS 1.2.
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -244,7 +244,7 @@ zeros
 
 ~~~
      struct {
-         ContentType special_type = tls12_cid; 
+         ContentType outer_type = tls12_cid; 
          ProtocolVersion version;
          uint16 epoch;
          uint48 sequence_number;
@@ -255,7 +255,7 @@ zeros
 ~~~~
 {: #dtls-ciphertext title="DTLS 1.2 CID-enhanced Ciphertext Record."}
 
-special_type
+outer_type
 :  The outer content type of a DTLSCiphertext record carrying a CID
    is always set to tls12_cid(TBD2). The real content
    type of the record is found in DTLSInnerPlaintext.real_type after
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -138,7 +138,7 @@ receive as a CID in encrypted records, it is possible
 for an endpoint to use a globally constant length for such connection
 identifiers.  This can in turn ease parsing and connection lookup,
 for example by having the length in question be a compile-time constant.
-Such implementations must still be able to send
+Such implementations MUST still be able to send
 CIDs of different length to other parties.
 Implementations that want to use variable-length CIDs are responsible
 for constructing the CID in such a way that its length can be determined
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -263,7 +263,13 @@ outer_type
 
 cid
 :  The CID value, cid_length bytes long, as agreed at the time the extension
-   has been negotiated.
+   has been negotiated.  Recall that (as discussed previously) each peer chooses
+   the CID value it will receive and use to identify the connection, so an
+   implementation can choose to always recieve CIDs of a fixed length.  If,
+   however, an implementation chooses to receive different lengths of CID,
+   the assigned CID values must be self-deliniating since there is no other
+   mechanism available to determine what connection (and thus, what CID length)
+   is in use.
 
 enc_content
 :  The encrypted form of the serialized DTLSInnerPlaintext structure.
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -69,7 +69,7 @@ receiver will be unable to locate the correct security context.
 
 #  Introduction
 
-The Datagram Transport Layer Security (DTLS) protocol was designed for
+The Datagram Transport Layer Security (DTLS) {{RFC6347}} protocol was designed for
 securing connection-less transports, like UDP. DTLS, like TLS, starts
 with a handshake, which can be computationally demanding (particularly
 when public key cryptography is used). After a successful handshake,
@@ -77,11 +77,11 @@ symmetric key cryptography is used to apply data origin
 authentication, integrity and confidentiality protection. This
 two-step approach allows endpoints to amortize the cost of the initial
 handshake across subsequent application data protection. Ideally, the
-second phase where application data is protected lasts over a longer
+second phase where application data is protected lasts over a long
 period of time since the established keys will only need to be updated
 once the key lifetime expires.
 
-In the current version of DTLS, the IP address and port of the peer are used to
+In DTLS as specified in RFC 6347, the IP address and port of the peer are used to
 identify the DTLS association. Unfortunately, in some cases, such as NAT rebinding,
 these values are insufficient. This is a particular issue in the Internet of Things
 when devices enter extended sleep periods to increase their battery lifetime. The
@@ -138,10 +138,11 @@ receive as a CID in encrypted records, it is possible
 for an endpoint to use a globally constant length for such connection
 identifiers.  This can in turn ease parsing and connection lookup,
 for example by having the length in question be a compile-time constant.
-Implementations, which want to use variable-length CIDs, are responsible
+Such implementations must still be able to send
+CIDs of different length to other parties.
+Implementations that want to use variable-length CIDs are responsible
 for constructing the CID in such a way that its length can be determined
-on reception. Such implementations must still be able to send
-CIDs of different length to other parties. Note that there is no CID 
+on reception.  Note that there is no CID
 length information included in the record itself.
 
 In DTLS 1.2, CIDs are exchanged at the beginning of the DTLS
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -138,10 +138,11 @@ receive as a CID in encrypted records, it is possible
 for an endpoint to use a globally constant length for such connection
 identifiers.  This can in turn ease parsing and connection lookup,
 for example by having the length in question be a compile-time constant.
-Implementations, which want to use variable-length CIDs, are responsible
+Such implementations must still be able to send
+CIDs of different length to other parties.
```
boaks, line 142: That's an interesting point. some weeks ago I considered to raise an issue for that. Generally constraint peers don't limit the CID length just for fun. They are not able to use larger ones. The upper limit of 255 bytes (hope I'm right) is more academical value. I guess, in constraint deployments 32 bytes will be mostly enough. My conclusion was therefore, that a server just documents his max. CID length. If more flexibility is really required, but constraint devices should be supported, I guess that max cid-length must be negotiated.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -370,25 +377,30 @@ receive and process DTLS records. No such test is defined in this	specification.
...
 Section 4.1.2.6 of DTLS 1.2. Both solutions, the "anti-replay window" or 
-"newer algorithm" will prevent address updates from replay attacks while the 
+"newer" algorithm, will prevent address updates from replay attacks while the 
 latter will only apply to peer address updates and the former applies to any 
 application layer traffic.
 
+Note that datagrams that pass the DTLS cryptographic verification procedures
+but do not trigger a change of peer address are still valid DTLS records and
+are still to be passed to the application.
+
```
boaks, line 389: Great clarification note!

draft-ietf-tls-dtls-connection-id.md
kaduk: The intent here was just to improve the readability of the transition from "when/how you can use fixed CID length" to "considerations about variable-length CIDs"; the actual technical requirements are specified elsewhere in the document.
(I think I also asked a question about whether there would be reason to lower the 255-byte limit, in the emailed review comments, which might be a better place to discuss this topic.)

It seems that the scenario where a node can only handle max (say) 32 byte CID is asymmetric if that node is client or server.  If it's the server, it has the option of rejecting CID entirely if the client uses a longer CID than it supports, but the client would have to do a fallback dance (retry without the CID extension) if the server said to use a "too-long" CID.  I was reluctant to suggest a very small cap on CID length since (IIRC) some people wanted to use the CID value as a self-encrypted blob to store routing information/etc.

ekr: FWIW, QUIC caps it at 20 octets, but does not allow the peer to indicate a restriction.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -254,15 +255,21 @@ zeros
...
    decryption.
 
 cid
 :  The CID value, cid_length bytes long, as agreed at the time the extension
-   has been negotiated.
+   has been negotiated.  Recall that (as discussed previously) each peer chooses
+   the CID value it will receive and use to identify the connection, so an
+   implementation can choose to always recieve CIDs of a fixed length.  If,
+   however, an implementation chooses to receive different lengths of CID,
+   the assigned CID values must be self-deliniating since there is no other
```
```suggestion
   the assigned CID values must be self-delineating since there is no other
```

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -486,7 +497,8 @@ that is malicious. This attack is of concern when there is a large asymmetry
 of request/response message sizes. 
 
 Additionally, an attacker able to observe the data traffic exchanged between 
-two DTLS peers is able to replay datagrams with modified IP address/port numbers. 
+two DTLS peers is able to replay datagrams with modified IP address/port numbers
+if the optional DTLS replay detection functionality is not in use.
```
ekr: I think it would be better to just note that you can replace those packets. the replay point seems like a distraction.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -506,8 +519,8 @@ TBD1    connection_id   -        Y          N           [[This doc]]
 
 Note: The value "N" in the Recommended column is set because this 
 extension is intended only for specific use cases. This document describes 
-an extension for DTLS 1.2 only; it is not to TLS (1.3). The DTLS 1.3 
-functionality is described in {{I-D.ietf-tls-dtls13}}. 
+an extension for DTLS 1.2 only; it is not applicable to TLS or DTLS 1.3.
+Analogous functionality for DTLS 1.3 is described in {{I-D.ietf-tls-dtls13}}.
```
ekr: This text actually seems like it's OBE, as DTLS 1.3 uses the same extension.

ekr (changes requested):

draft-ietf-tls-dtls-connection-id.md
kaduk: The replays will be accepted as valid, though, whereas arbitrary replacement would not.
Or are you saying that those *fields* can be replaced, not the packets themselves...

ekr: what I'm saying is that you could have suppressed the packets and played them with your own IP and that attack would be just as good and isn't prevented by the DTLS anti-replay mechanism.

boaks: > played them with your own IP

That "own IP" isn't used for too much, potentially to send a message back.
With "IP changes only on newer messages (according the record sequence number)", that narrows the attack to even that "newest package once", until an other message gets the newest (including the spoofed message itself). 
I tried to express my assumption about that attack in my comment to PR #71 .
For me it seems to be not too likely, that someone tries to DDoS using something, waiting passive for passing messages instead of active generating messages. That's mainly a question of the assumed message volume.

kaduk: I think the goal here is to describe what can be done; there may not need to be a great motivation for or benefit from doing so.

I took a harder think about about this and pushed some new text.  I think that it is still worth talking about the replay case, since the replays can be modified as well, but the straight-up modification case is more promiinent.

draft-ietf-tls-dtls-connection-id.md
kaduk: Some further updates pushed here.  It seems simplest to put the 'CH,SH' in the TLS 1.3 column in this document rather than have dtls13 modify the registration, but there's also a decent argument for keeping all the 1.3 stuff out of this document.  (So I made it as a separate commit that is easy to remove if needed.)

draft-ietf-tls-dtls-connection-id.md
boaks: Let me first mention, that both, the new text and ekr's comment are pretty right.

But since more than a year it seems to be still unclear, if, beside an "amplification attack", that causes any more trouble? So, yes, that is a possible scenario, that someone is changing the ip-address. But with that, what will be the negative effect of that? The record will be processed and only the usage of that spoofed address to send something back seems to be the critical part.

FMPOV, it' s very easy:
if the ip-address is moved out of cryptographic validation with the CID,
-  then this is either very general a "no go", because it's not possible to limit the damage of it
- or it is possible to define the limits of that attacks.

So, is this going to define the limits? Or should we consider, that, if the ip-address is outside of the cryptographic validation, it's not possible to define that?

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -486,7 +497,11 @@ that is malicious. This attack is of concern when there is a large asymmetry
 of request/response message sizes. 
 
 Additionally, an attacker able to observe the data traffic exchanged between 
-two DTLS peers is able to replay datagrams with modified IP address/port numbers. 
+two DTLS peers is able to modify IP address/port numbers without causing the
+packet to be rejected by the recipient.  When the optional DTLS replay
+detection is not in use, such an attacker can also replay the observed packets,
```
boaks, line 501: I'm not sure. 
For me "replay protection" requires more than the "newer filter". But the "newer filter" would already provide a level of protection, which is comparable with the "replay protection".

draft-ietf-tls-dtls-connection-id.md
boaks: > what I'm saying is that you could have suppressed the packets and played them with your own IP and that attack would be just as good and isn't prevented by the DTLS anti-replay mechanism.

Is the intention of this clarification, that the simple-cases, where either a "replay-protection" or a "newer" may help, should be removed and the "address validation" should be accented?

kaduk: My main goal here is to accentuate aspects that are different and possibly "surprising" than regular DTLS due to the addition of the CID (and no longer relying on 5-tuple).  In regular DTLS the attacker can't modify the IP addresses because they're a lookup key, so the ability to do that at all is noteworthy.  Likewise, for regular DTLS (without replay protection) an attacker can replay many copies, but the addresses in them (and thus the entire packet itself) have to be the same every time.  With CIDs, the replayed copies don't have to be bit-for-bit identical, which also seems noteworthy -- it's a "malleable" replay.  The specific consequences of this will vary from case to case, and the one "amplification" attack that we've identified so far is particularly frightening (and we have some partial mitigations for), but we should still state the new properties since we are not considering all potential use cases.  In other words, the goal here is not to specify exactly all sorts of "trouble" this can cause, but rather to note the difference from the previous state of affairs, so that the reader can think about how (if at all) it affects their use case.  With that mindset, I do not find the question of whether it is possible to define the "limits of that attacks" to be worth pursuing very much.

draft-ietf-tls-dtls-connection-id.md
kaduk: The replay protection has to handle out-of-order delivery, usually with a sliding window that tracks whether the most recent N packets have been seen or not.  I'm not sure I understand your question here, though.

boaks: Try to precise my saying:

The general issue of "replay" is **not** introduced by CID. The general protection against such "replay" is already in RFC6347. I would prefer, if this draft focus on the changes and **new** threats caused by CID and not repeat the ones, which apply also without CID.

draft-ietf-tls-dtls-connection-id.md
boaks: > With CIDs, the replayed copies don't have to be bit-for-bit identical, which also seems noteworthy

Example?

boaks: > In other words, the goal here is not to specify exactly all sorts of "trouble" this can cause, but rather to note the difference from the previous state of affairs, so that the reader can think about how (if at all) it affects their use case. 

> With that mindset, I do not find the question of whether it is possible to define the "limits of that attacks" to be worth pursuing very much.

I would prefer, if the first is discussed in the community, so that everybody can benefit from the accumulated result. That requires on my view, to address or discuss the considered "limits of that attacks".

draft-ietf-tls-dtls-connection-id.md
boaks: Let me add:

> is able to modify IP address/port numbers without causing the packet to be rejected by the recipient.

That seems to be generally true, even without CID. Otherwise (s)NATs  will not work.
What is introduced with CID, is that a **different** modification (different from such a potential modification during the handshake), may be applied. CID is especially made for that scenario. And there are no direct means to precise distinguish such assumed changes from those, which are applied by an attacker. At least not without an additional data exchange. There may be some scenarios considered to be more likely as an attack than others, but that are more assumptions.

boaks: > Additionally, an attacker able to observe the data traffic exchanged between two DTLS peers is able to replay datagrams with modified IP address/port numbers.

I understand two issues here:
- message replay (reprocess the received data). CID doesn't change that, so [RFC6347 Anti-Replay](https://tools.ietf.org/html/rfc6347#section-4.1.2.6) is still valid.
- address/port change. The effect depends on the [peer-address-update](https://github.com/tlswg/dtls-conn-id/blob/master/draft-ietf-tls-dtls-connection-id.md#peer-address-update-peer-address-update).

The only change I see, that it may be considered to be easier for an observer to replay the message with the **own address** compared to replay it spoofing the original address as source address.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: I am fine with the editorial change. As there is no desire to change implementation of the mechanism or to add new requirements I believe there is not much more to do with this item.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: I don't care about the direction we take here. If you think it is easier for IANA to register the value with this document then that's fine for me.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -138,10 +138,11 @@ receive as a CID in encrypted records, it is possible
 for an endpoint to use a globally constant length for such connection
 identifiers.  This can in turn ease parsing and connection lookup,
 for example by having the length in question be a compile-time constant.
-Implementations, which want to use variable-length CIDs, are responsible
+Such implementations must still be able to send
```
```suggestion
Such implementations MUST still be able to send
```

draft-ietf-tls-dtls-connection-id.md
ekr: I think there is a normative MUST here.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: From the discussions I wonder whether we should make this text change. It looks like it confuses more than it clarifies.

draft-ietf-tls-dtls-connection-id.md
hannestschofenig: MUST language added.

PR #77, opened 2020-11-25
Update to new (hopefully) injective MAC structure as discussed in the meeting and on-list

ekr:
I added a placeholder seq_num in order to put the tls12_cid in the right place. Note that I also did not add an iv_length. This is implicit in the ciphersuite selection and so you'd have to be making some quite strong assumptions to have the iv_length be necessary, it seems, and that's already the way things were *without* the CID, so presumably if those obtained you could still mount an attack.

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -335,7 +335,7 @@ connection ID. The former always consists of a sequence number
 followed by some other content type than tls12_cid; the latter
 always consists of the seq_num_placeholder followed by tls12_cid.
 Although 2^64-1 is potentially a valid sequence number, tls12_cid
-will never be a valid CID when the connection ID is not in use.
+will never be a valid content type when the connection ID is not in use.
 In addition, the epoch and sequence_number are now fed into
 the MAC in the same order as they appear on the wire.
 
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -367,7 +367,7 @@ data the following modification is made to the additional data calculation.
 
 ~~~
     additional_data = seq_num_placeholder +
-                      tls12_cid,
+                      tls12_cid +
                       cid_length +
                       tls12_cid +
                       DTLSCiphertext.version +
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -315,8 +315,8 @@ described in {{RFC7366}}.
 ~~~
     MAC(MAC_write_key,
         seq_num_placeholder +
-        tls12_cid,
-        cid_length,
+        tls12_cid +
+        cid_length +
         tls12_cid +
         DTLSCiphertext.version +
         epoch +
@@ -348,8 +348,8 @@ described in {{RFC7366}}.
 ~~~
     MAC(MAC_write_key,
         seq_num_placeholder +
-        tls12_cid,
-        cid_length,
+        tls12_cid +
+        cid_length +
         tls12_cid +
         DTLSCiphertext.version +
         epoch +
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -301,6 +301,9 @@ length_of_DTLSInnerPlaintext
 : The length (in bytes) of the serialised DTLSInnerPlaintext (two-byte integer).
   The length MUST NOT exceed 2^14.
 
+seq_num_placeholder
+: 8 bytes of 0xff
+
 Note "+" denotes concatenation.
 
 ## Block Ciphers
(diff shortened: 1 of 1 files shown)


ekr:
@kaduk WDYT?

kaduk (approved):
This looks good; thank you.
(I had a 'todo' note to check with you on who had the action-item here ... now I know!)

boaks:
In my opinion, there is no reason to use a fixed head with "0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0x19".

ekr:
Well, I think I've explained the rationale for it in the text. Do you believe that the claimed properties are incorrect or merely that they're not necessary?

boaks:
I understand your explanation to put 0xff 8 times in the front of the marker copy of tls_cid.  
Though that results in a fixed header of 9 bytes, I'm curious, why that should improve the MAC.
So back to my question before, why should there be the marker copy of the tls_cid in front at all?
It was said, that all versions use such marker (maybe that's also a misunderstanding). But at least, for 1.2, such markers wasn't the case.

So, the discussion starts with: 
Why should the `marker`, a copy of something, which is already in the MAC, be additionally at the begin?

Anyway, if the explanation gets too complex, and others don't feel curious as well, there is also no reason not to agree on that MAC definition of this PR; it could be obviously calculated.

boaks:
Let me add: I'm not too happy with this proposal. But I will be even more unhappy, if this turns into a endless discussion. I would prefer some signs from others (@thomas-fossati, @hannestschofenig), if they are OK with it. 
Not to start in some weeks again with discussing the MAC over and over ...

ekr:
The idea here is to provide domain separation between the MAC inputs when CID is in use and when it is not, for the reason Ben indicated, namely ensuring that an active attacker cannot create confusion about the inputs to the MAC.

When CID is not in use, the first bytes are the 8-byte epoch/seq and the content type. Thus, if we want to ensure that the input is different when in use, then we must have 8 placeholder bytes and a different content type (or some other fixed-position distinguisher).

boaks:
Isn't there is already a separator?
In my opinion, the record-type does already provide that!
All MAC starts with 8 bytes (epoch +seqn) followed by the record-type. Though that record-type of cid-records already differ from the other, non-cid, record-types, I simply don't see the benefit.
An non-cid application-record use 0x17 at the 9th byte, with cid an application-record use 0x19.

ekr:
Yes, that would be true if we weren't also moving the sequence number to make the DTLS header input into the MAC as-is rather than broken apart as it currently is. Once you've done that, then the type byte would be somewhere later in the packet -- in the version number if I am counting correctly.

boaks:
I start to understand:
In difference to RFC6347, now the header goes as on the wire into the MAC. 
For that, a artificial dummy header (fixed 9 byte) is put again in front of that, in order to ensure, that this differs from other records. Also as consequence, the cid-length is now between the dummy-header and the on-wire header.

thomas-fossati:
> I would prefer some signs from others (@thomas-fossati, @hannestschofenig), if they are OK with it.

LGTM

kaduk:
Good catch on the `+s not ,s`.  Still looks good.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -310,33 +313,51 @@ that do not use the with Encrypt-then-MAC processing
...
 ~~~
 
+The rationale behind this construction is to separate the MAC input
+for DTLS without the connection ID from the MAC input with the
+connection ID. The former always consists of a sequence number
+followed by some other content type than tls12_cid; the latter
+always consists of the seq_num_placeholder followed by tls12_cid.
+Although 2^64-1 is potentially a valid sequence number, tls12_cid
+will never be a valid CID when the connection ID is not in use.
+In addition, the epoch and sequence_number are now fed into
```
boaks, line 339: > will never be a valid CID when the connection ID is not in use.

I think, this should be:

> tls12_cid will never be one of the other content types when the connection ID is not in use.

The other MAC (when connection ID is not in use) uses the MAC of RFC5246 

```
MAC(MAC_write_key, seq_num +
                            TLSCompressed.type +
                            TLSCompressed.version +
                            TLSCompressed.length +
                            TLSCompressed.fragment);
```

not the one previously defined for this extension.

ekr: I'm not sure I follow the distinction you are trying to draw.

kaduk: I think we're looking for s/never be a valid CID/never be a valid content type/

ekr:
Duh. One more piece of evidence that it's hard to edit yourself. Updated.

draft-ietf-tls-dtls-connection-id.md
boaks: > I think we're looking for s/never be a valid CID/never be a valid content type/

Yes.

ekr: Thanks. Fixed.

chris-wood (approved):

chris-wood:
@seanturner this LGTM. Let's run this past the list!

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -345,11 +366,14 @@ For ciphers utilizing authenticated encryption with additional
 data the following modification is made to the additional data calculation.
 
 ~~~
-    additional_data = seq_num +
+    additional_data = seq_num_placeholder +
+                      tls12_cid,
```
hannestschofenig:
```suggestion
                      tls12_cid +
```

hannestschofenig:
I am OK with the changes. It is time to wrap this work up.

PR #78, opened 2020-12-07
It is no longer true that an empty CID and no CID are the same

ekr:


chris-wood (approved):
Pending @hannestschofenig's approval, too.

hannestschofenig (approved):

PR #82, opened 2021-02-02
Changed DTLSCiphertext to DTLSCIDCiphertext for easier readability

hannestschofenig:
Doing so I introduce ciphertext_length, as the length of the AEAD-encrypted form of the serialized DTLSInnerPlaintext structure.

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -216,8 +216,8 @@ When CIDs are being used, the content to be sent
 is first wrapped along with its content type and optional padding into a
 DTLSInnerPlaintext structure. This newly introduced structure is shown in
 {{dtls-innerplaintext}}. The DTLSInnerPlaintext
-byte sequence is then encrypted. To create the DTLSCiphertext structure shown in
-{{dtls-ciphertext}} the CID is added.
+byte sequence is then encrypted. To create the DTLSCIDCiphertext structure, shown in
+{{dtls-ciphertext}}, the CID is added.
 
 ~~~
      struct {
@@ -250,13 +250,13 @@ zeros
          uint48 sequence_number;
          opaque cid[cid_length];               // New field
          uint16 length;
-         opaque enc_content[DTLSCiphertext.length];
-     } DTLSCiphertext;
+         opaque enc_content[ciphertext_length];
+     } DTLSCIDCiphertext;
 ~~~~
 {: #dtls-ciphertext title="DTLS 1.2 CID-enhanced Ciphertext Record."}
 
 outer_type
-:  The outer content type of a DTLSCiphertext record carrying a CID
+:  The outer content type of a DTLSCIDCiphertext record carrying a CID
    is always set to tls12_cid(TBD2). The real content
    type of the record is found in DTLSInnerPlaintext.real_type after
    decryption.
@@ -272,7 +272,8 @@ cid
    is in use.
 
 enc_content
-:  The encrypted form of the serialized DTLSInnerPlaintext structure.
+:  The AEAD-encrypted form of the serialized DTLSInnerPlaintext structure of 
+   length ciphertext_length. 
 
 All other fields are as defined in RFC 6347.
 
@@ -317,7 +318,7 @@ described in {{RFC7366}}.
         tls12_cid +
         cid_length +
         tls12_cid +
-        DTLSCiphertext.version +
+        DTLSCIDCiphertext.version +
         epoch +
         sequence_number +
         cid +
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -250,13 +250,13 @@ zeros
          uint48 sequence_number;
          opaque cid[cid_length];               // New field
          uint16 length;
-         opaque enc_content[DTLSCiphertext.length];
-     } DTLSCiphertext;
+         opaque enc_content[ciphertext_length];
+     } DTLSCIDCiphertext;
```
boaks, line 254: I may be wrong.
For me, the old defines a `DTLSCiphertext`, which contains a `length` field, which is referenced by `DTLSCiphertext.length`. the new one now has still that field, but introduces a "ciphertext_length",  which may be considered as "maverick".

boaks: Would `DTLSCIDCiphertext.length` be a replacement for "ciphertext_length"?

hannestschofenig: Using DTLSCIDCiphertext.length is IMHO incorrect because the length of the encrypted content is different from the length of the entire DTLSCIDCiphertext structure. Do I see that wrong?

boaks: My understanding:
`DTLSCIDCiphertext.length` is not the length of `DTLSCIDCiphertext`, it's the value of the field `uint16 length` of that structure. 

e.g. 

```
   uint16 encLength;
   opaque enc_content[DTLSCIDCiphertext.encLength];
} DTLSCIDCiphertext;
```

[RFC5246](https://tools.ietf.org/html/rfc5246#page-21) contains

```
 struct {
          ContentType type;       /* same as TLSPlaintext.type */
          ProtocolVersion version;/* same as TLSPlaintext.version */
          uint16 length;
          opaque fragment[TLSCompressed.length];
      } TLSCompressed;
```

the `TLSCompressed.length` can't be the length of the whole `TLSCompressed` structure, but instead the value of the field `uint16 length`.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -272,7 +272,8 @@ cid
    is in use.
 
 enc_content
-:  The encrypted form of the serialized DTLSInnerPlaintext structure.
+:  The AEAD-encrypted form of the serialized DTLSInnerPlaintext structure of 
+   length ciphertext_length.
```
jsalowey, line 276: Why is this just for AEAD cipher types?

boaks: @jsalowey 

FMPOV, there are currently two interpretations of that definition. 

Hannes:
> Using DTLSCIDCiphertext.length is IMHO incorrect because the length of the encrypted content is different from the length of the entire DTLSCIDCiphertext structure. Do I see that wrong?

So xxx.length is the length of xxx.

My understanding:
> DTLSCIDCiphertext.length is not the length of DTLSCIDCiphertext, it's the value of the field uint16 length of that structure.

So xxx.length is the value of the field xxx.length.

I think, my interpretation is right.

If I'm right, then the changes for "easier readability" will change the term (field name) `length` in more places, not only there.

```sh
   ...
    uint16 ciphertext_length;
   opaque enc_content[DTLSCIDCiphertext.ciphertext_length];
} DTLSCIDCiphertext;

```

with that, also the term `length`in the text, not only in the structure, must be exchanged with `ciphertext_length`. And, of course, for the other cipher types as well.

With that, I'm not sure, if exchange that term really pays off, this style/expression is used in many other RFCs, and so should be common to the most readers.

(I guess, if @hannestschofenig finds the time to revisit this, he will agree on my interpretation and with that, it may either "change much more", or this PR may get withdrawn.)

draft-ietf-tls-dtls-connection-id.md
ekr: Yes, this interpretation is correct.

ekr:
It seems to me that there are two changes here:

1. The change to DTLSCIDCiphertext -- innocuous but maybe unnecessary
2. The change to encrypted_record, which is, I believe, wrong. You'll note that RFC8446 uses the style pre-PR.

I suggest we either close this PR or just do 1.

boaks:
Just for information:

I assume, that the irritation about `DTLSCiphertext.length` is mainly caused by very common implicit definitions of `???.length` in in programing  languages, e.g. Java, where

```java
byte[] data = new byte[0124];
...
if (data.length > 0) ...
...
``` 

To escape such accidentally misinterpretations using a name different from "length" (and please not "size" nor "sizeof") may help. But I have my doubts, if the effect is worth the work. And potentially, all others, which are common to the notation within the RFC may then get mixed up by such new names.

hannestschofenig:
I have been trying to address the review from Ben and, reading through your comments, my attempt to improve the readability failed. This means that I will withdraw this PR.

PR #83, opened 2021-02-02
Added reference to draft-gont-numeric-ids-sec-considerations

hannestschofenig:


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -525,6 +526,8 @@ of request/response message sizes.
 
 Additionally, an attacker able to observe the data traffic exchanged between
 two DTLS peers is able to replay datagrams with modified IP address/port numbers.
+Guidelines for generating transient identifiers, like port numbers and CIDs, can 
+be found in {{I-D.gont-numeric-ids-sec-considerations}}.
```
boaks, line 530: I'm not sure, why `port numbers` are mentioned. Any example related to CID?

ekr: This seems both unnecessary and unhelpful, for the reasons listed in my review of draft-gont. I do not believe we should make this change.

jsalowey: If the authors are not in consensus to add this then we should not add it.

PR #92, opened 2021-03-12
Adding reference to RFC 8174

hannestschofenig:


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -101,7 +102,8 @@ extension.
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
```
thomas-fossati, line 104: This whole para could be replaced by a `{::boilerplate bcp14}`

kaduk: Probably better to do it that way, yes, though what's in the PR ought to work as well.

kaduk (approved):

PR #96, opened 2021-03-14
zero-length CID clarification

### LINKED ISSUE #89: Zero-length CID clarifications
Tom Petch's review: https://mailarchive.ietf.org/arch/msg/tls/Opmv1JCmY10uBKP1SrzfUpxhzCY/

```
s.3
I found this unclear until I had understood it all (or perhaps I do not
understand it)

"...or again, alternately, to use a zero-length CID)."
This suggests that a zero length CID is valid in Application Data which
later text seems to contradict; otherwise I cannot see what this is saying.

"  If DTLS peers have negotiated the use of a CIDs using the ClientHello
and the ServerHello messages "
arguably sending a zero CID and receiving a zero CID is a successful
Hello negotiation perhaps
" If DTLS peers have negotiated the use of a non-zero CID in at least
one direction, using the ClientHello and the ServerHello messages"

"The DTLS peers determine whether incoming and outgoing messages need.."
seems not to cater for unidirectional CIDs; perhaps
"The DTLS peers determine whether incoming or outgoing, or both,
messages need.. "
```

thomas-fossati:
Clarify that successfully negotiating two zero-len CIDs is the same as
failing the negotiation in terms of what happens on the wire afterwards.

Fixes #89

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -160,7 +160,7 @@ If DTLS peers have not negotiated the use of CIDs, which includes the case
 where both sent a zero-length cid in their connection_id extensions, then the
 RFC 6347-defined record format and content type MUST be used.
 
-If DTLS peers have negotiated the use of a non-zero CID in at least one
+If DTLS peers have negotiated the use of a non-zero-length CID in at least one
 direction using the ClientHello and the ServerHello messages, then the peers
 need to take the following steps.
 
@@ -678,4 +678,3 @@ better privacy features and developers are encouraged to switch to the new versi
 
 Finally, we want to thank the IETF TLS working group chairs, Chris Wood, Joseph Salowey, and
 Sean Turner, for their patience, support and feedback.
-
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -135,8 +135,7 @@ A server willing to use CIDs will respond with a "connection_id"
 extension in the ServerHello, containing the CID it wishes the
 client to use when sending messages towards it. A zero-length value
 indicates that the server will send with the client's CID but does not
-wish the client to include a CID (or again, alternately, to use a
-zero-length CID).
+wish the client to include a CID.
 
 Because each party sends the value in the "connection_id" extension it wants to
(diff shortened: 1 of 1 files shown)


boaks:
For me, this PR makes it clearer and better understandable.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -157,13 +156,15 @@ DTLS 1.2 in general does not allow TLS 1.3-style post-handshake messages
...
-the ServerHello messages then the peers need to take the following steps.
+If DTLS peers have negotiated the use of a non-zero CID in at least one
+direction using the ClientHello and the ServerHello messages, then the peers
+need to take the following steps.
 
-The DTLS peers determine whether incoming and outgoing messages need
+The DTLS peers determine whether incoming or outgoing (or both) messages need
 to use the new record format, i.e., the record format containing the CID.
 The new record format with the the tls12_cid content type is only used once encryption
 is enabled. Plaintext payloads never use the new record type and the CID content
```
kaduk, line 170: A few lines below, we say "For sending, if a zero-length CID has been negotiated" which is perhaps in conflict with our stance on "negotiated" after the rest of these changes.  Perhaps "if a zero-length CID was advertised for this direction"?

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -157,13 +156,15 @@ DTLS 1.2 in general does not allow TLS 1.3-style post-handshake messages
...
 
-If DTLS peers have not negotiated the use of CIDs then the RFC 6347-defined
-record format and content type MUST be used.
+If DTLS peers have not negotiated the use of CIDs, which includes the case
+where both sent a zero-length cid in their connection_id extensions, then the
+RFC 6347-defined record format and content type MUST be used.
 
-If DTLS peers have negotiated the use of a CIDs using the ClientHello and
-the ServerHello messages then the peers need to take the following steps.
+If DTLS peers have negotiated the use of a non-zero CID in at least one
```
kaduk:
```suggestion
If DTLS peers have negotiated the use of a non-zero-length CID in at least one
```

kaduk (approved):
What's here is good, but see inline on whether we should do a little more.

PR #100, opened 2021-04-14
Try to clean up the text a bit

ekr:


ekr:
@kaduk

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -154,32 +154,24 @@ DTLS 1.2 in general does not allow TLS 1.3-style post-handshake messages
...
-need to take the following steps.
-
-The DTLS peers determine whether incoming or outgoing (or both) messages need
-to use the new record format, i.e., the record format containing the CID.
-The new record format with the the tls12_cid content type is only used once encryption
-is enabled. Plaintext payloads never use the new record type and the CID content
+If DTLS peers have not negotiated the use of CIDs, or a zero-length
+CID has been advertised for a given direction, then the RFC
+6347-defined record format and content type MUST be used to send in
+that direction.
```
kaduk: (nit) something of a singular/plural mismatch since "not negotiated the use of CIDs" doesn't have an associated direction.
On first look, "MUST be used to send in the indicated direction(s)" might cover it.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -154,32 +154,24 @@ DTLS 1.2 in general does not allow TLS 1.3-style post-handshake messages
...
-the connection and the security association. If the tls12_cid content type is not set,
-then the connection and security association is looked up by the 5-tuple and a
-check MUST be made to determine whether the expected CID value is indeed
-zero length. If the check fails, then the datagram MUST be dropped.
+When receiving, if the tls12_cid content type is set, then the CID is
+used to look up the connection and the security association. If the
+tls12_cid content type is not set, then the connection and security
+association is looked up by the 5-tuple and a check MUST be made to
+determine whether a non-zero length CID is expected. If the check
+fails, then the datagram MUST be dropped.
```
kaduk: "the check fails" is maybe ambiguous about how to map from "a check [...] whether <X> is expected".  Maybe just "If a non-zero-length CID is expected for the retrieved association, then the datagram MUST be dropped"?

kaduk (commented):
The overall approach is a good cleanup; just some wordsmithing left.

ekr:
@kaduk I cleaned this up a bit, PTAL.

kaduk (approved):
Updates look good.  Thanks!

PR #101, opened 2021-04-14
IANA updates. Fixes #99.

### LINKED ISSUE #99: new "connection_id" extension codepoint after MAC semantics change?
We currently have an early allocation of the TLS extensiontype codepoint 53 that was allocated on 2019-07-02.
An implementation of the draft at that time will no longer interoperate with the current specification due to changes in how the MAC inputs and AAD are laid out.

If we want to allocate a new extensiontype value for the new semantics, we should tell IANA that there is an existing early allocation and, e.g., that it should be left to expire (and what version of the draft to use as a reference until it does), with a new codepoint to be allocated in its stead.

We also have the option to direct IANA that "once this document is approved for publication, the early allocation will be deprecated and the following assignment will be made".

ekr:
- Suggest a column name.
- Deprecate the old connection_id codepoint.

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -534,11 +534,11 @@ The topic of peer address updates is discussed in {{peer-address-update}}.
 
 #  IANA Considerations
 
-This document requests three actions by IANA. 
+This document requests three actions by IANA.
 
 ## Extra Column to TLS ExtensionType Values Registry
 
-IANA is requested to add an extra column to the
+IANA is requested to add an extra column named "DTLS-Only" to the
 "TLS ExtensionType Values" registry to indicate whether an extension is only
 applicable to DTLS and to include this document as an additional reference
 for the registry.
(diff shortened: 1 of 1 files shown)


ekr:
@kaduk

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -534,11 +534,11 @@ The topic of peer address updates is discussed in {{peer-address-update}}.
...
-This document requests three actions by IANA. 
+This document requests three actions by IANA.
 
 ## Extra Column to TLS ExtensionType Values Registry
 
-IANA is requested to add an extra column to the
+IANA is requested to add an extra column named "DTLS-Only" to the
 "TLS ExtensionType Values" registry to indicate whether an extension is only
 applicable to DTLS and to include this document as an additional reference
 for the registry.
```
kaduk, line 544: IIRC in the thread with IANA we did state the expectations for how this column is populated (Y and N), so I think it's okay to not say more here.

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -547,7 +547,10 @@ for the registry.
 
 IANA is requested to allocate an entry to the existing "TLS ExtensionType
 Values" registry, for connection_id(TBD1) as described
-in the table below. 
+in the table below. Although there is an early allocation, it
```
kaduk:
```suggestion
in the table below. Although the value 53 has been allocated by early allocation for a previous version of this document, it
```

kaduk (approved):
See suggestion inline, though even as-is this is probably enough to cause the right things to happen at IANA and the RPC.

PR #102, opened 2021-04-14
fix nit

kaduk:
It is a new record format; the new type is the CID content type mentioned
shortly thereafter.


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -163,7 +163,7 @@ If DTLS peers have negotiated the use of a non-zero-length CID for a
 given direction, then once encryption is enabled they MUST send with
 the record format defined in {{dtls-ciphertext} with the
 new MAC computation defined in {{mac}} and the content type tls12_cid.
-Plaintext payloads never use the new record type and the CID content
+Plaintext payloads never use the new record format and the CID content
```
```suggestion
Plaintext payloads never use the new record format or the CID content
```

kaduk: good catch; thanks.

Issue #104, opened 2021-04-19
Rob Wilton's IESG review


thomas-fossati:
https://mailarchive.ietf.org/arch/msg/tls/ssfL3x2jgoHWOx2RtK0033_V4K4/

```
Hi,

I'm no DTLS expert, but I found the concepts/explanation in this document easy
to follow.

I was slightly confused by the requirement to encode the length in variable
length CIDs, and had to read the relevant text a second time.   As a
suggestion, it might help if these two sentences were reworded the other way
round:

OLD:
Implementations that want to use
   variable-length CIDs are responsible for constructing the CID in such
   a way that its length can be determined on reception.  Note that
   there is no CID length information included in the record itself.

NEW:
Since the CID length information is not included in the record itself,
implementations that want to use ... <as before>.

One minor question.  In the contributors, I noted that Jana is listed as being
associated with Google, but it might be worth checking if that is still
accurate.

Regards.
Rob
```

thomas-fossati:
Regarding:

> One minor question.  In the contributors, I noted that Jana is listed as being associated with Google, but it might be worth checking if that is still accurate.

We could streamline that a bit:
1. removing the affiliations (who knows which have become stale in the meantime?), and also
2. `strings.Join(cid_task_force_members, ",")`

Besides, is "task force" the right term?  Would "QUIC CID design team" be a better designation?

PR #105, opened 2021-04-19
rephrasing the "cid with encoded length" para

thomas-fossati:
Partly addresses #104


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -142,10 +142,10 @@ identifiers.  This can in turn ease parsing and connection lookup,
 for example by having the length in question be a compile-time constant.
 Such implementations MUST still be able to send
 CIDs of different length to other parties.
-Implementations that want to use variable-length CIDs are responsible
+Since the CID length information is not included in the record itself,
+implementations that want to use variable-length   CIDs are responsible
```
thomas-fossati:
```suggestion
implementations that want to use variable-length CIDs are responsible
```

PR #111, opened 2021-04-21
John Scudder's Improved Peer Address Update Text

hannestschofenig:


draft-ietf-tls-dtls-connection-id.md
```diff
@@ -409,8 +410,8 @@ Note that datagrams that pass the DTLS cryptographic verification procedures
 but do not trigger a change of peer address are still valid DTLS records and
 are still to be passed to the application.
 
-Application protocols that implement protection against these attacks depend on
-being aware of changes in peer addresses so that they can engage the necessary
+(*) Note: Application protocols that implement protection against spoofed addresses
```
thomas-fossati, line 413: I would stick an informational ref to {{I-D.ietf-core-echo-request-tag}} here.

thomas-fossati: (For the records: this is just a suggestion, not a blocking comment.  Feel free to ignore it.)

PR #116, opened 2021-04-23
Lars' grammar fixes

### LINKED ISSUE #115: Lars Eggert's IESG review
https://mailarchive.ietf.org/arch/msg/tls/Vg7QymL8fu-pAHw9Bs4CD46KC_U/

```
All comments below are about potential very minor issues that you may choose to
address in some way - or ignore - as you see fit. Some were flagged by
automated tools, so there will likely be some false positives. There is no need
to let me know what you did with these suggestions. 

Section 3, paragraph 10, nit:
>    the record format defined in {{dtls-ciphertext} with the new MAC

Broken kramdown reference?

Section 1, paragraph 4, nit:
-    This document defines an extension to DTLS 1.2 to add a CID to the
+    This document defines an extension to DTLS 1.2 to add a Connection ID (CID) to the
+                                                             ++++++++++  ++++++

Section 3, paragraph 7, nit:
-    for example by having the length in question be a compile-time
+    for example, by having the length in question be a compile-time
+               +

Section 3, paragraph 7, nit:
-    different length to other parties.  Implementations that want to use
-    variable-length CIDs are responsible for constructing the CID in such
+    different lengths to other parties.  Implementations that want to use
+                    +
+    variable-length CIDs are responsible for constructing the CIDs in such
+                                                                 +

Section 3, paragraph 12, nit:
-    datagram with the RFC 6347-defined record format the MAC calculation
+    datagram with the RFC 6347-defined record format, the MAC calculation
+                                                    +

Section 4, paragraph 6, nit:
-    *  The true content type is inside the encryption envelope, as
-           - -
+    *  The real content type is inside the encryption envelope, as
+             ++

Section 6, paragraph 2, nit:
-    datagram unless the following three conditions are met:
+    datagram, unless all of the following three conditions are met:
+            +       +++++++

Section 10, paragraph 2, nit:
-    This document requests three actions by IANA.
-                                         ^^
+    This document requests three actions from IANA.
+                                         ^^^^

Section 4, paragraph 17, nit:
> cord. outer_type The outer content type of a DTLSCiphertext record carrying a 
>                                    ^^^^^^^^^
If 'type' is a classification term, 'a' is not necessary. Use "type of". (The
phrases 'kind of' and 'sort of' are informal if they mean 'to some extent'.)

Section 4, paragraph 19, nit:
> the CID value it will receive and use to identify the connection, so an implemen
>                                   ^^^^^^
Make sure that 'use to' is correct. For habitual actions in the past or to mean
'accustomed to', use "used to".

Section 5, paragraph 6, nit:
> Plaintext The length (in bytes) of the serialised DTLSInnerPlaintext (two-byte inte
>                                        ^^^^^^^^^^
Do not mix variants of the same word ('serialise' and 'serialize') within a
single text.

Section 6, paragraph 2, nit:
>  that has a source address different than the one currently associated with the D
>                                      ^^^^
Did you mean 'different "from"? 'Different than' is often considered colloquial
style.

Section 6, paragraph 2, nit:
> ied in the received datagram, unless all of the following three conditions are met: 
>                                      ^^^^^^^^^^
Consider using "all the".

"Appendix A.", paragraph 1, nit:
>  History RFC EDITOR: PLEASE REMOVE THE THIS SECTION draft-ietf-tls-dtls-connect
>                                    ^^^^^^^^
Maybe you need to remove the second determiner so that only "THE" or "THIS" is
left.

"Appendix A.", paragraph 29, nit:
> 2 * Move to internal content types a la DTLS 1.3. draft-ietf-tls-dtls-conne
>                                    ^^^^
'a la' is an imported foreign expression, which originally has a diacritic.
Consider using "à la"

"Appendix B.", paragraph 1, nit:
> formation RFC EDITOR: PLEASE REMOVE THE THIS SECTION The discussion list for the
>                                     ^^^^^^^^
Maybe you need to remove the second determiner so that only "THE" or "THIS" is
left.

"Appendix C.", paragraph 1, nit:
>  Many people have contributed to this specification and we would like to thank the following
>                                       ^^^^^^^^^^^^^^^^^
Use a comma before 'and' if it connects two independent clauses (unless they
are closely connected and short).

These reference issues exist in the document:
 * No reference entries found for: 
     [ChangeCipherSpec], [length], [length_of_padding],
     [DTLSCiphertext.length],
     [draft-rescorla-tls-dtls-connection-id-00],
     [cid_length], [DTLSPlaintext.length]
 * Uncited references: [RFC5246]
 * Obsolete reference to RFC5246, obsoleted by RFC8446

These URLs in the document did not return content:
 * https://www1.ietf.org/mailman/listinfo/tls
 * http://www.ietf.org/internet-drafts/draft-tschofenig-tls-dtls-rrc-01.txt
 * http://www.ietf.org/internet-drafts/draft-ietf-tls-dtls13-40.txt
```

thomas-fossati:
Fixes #115

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -49,7 +49,6 @@ author:
 
 normative:
   RFC2119:
-  RFC5246:
   RFC6347:
   RFC8446:
 
@@ -177,7 +176,7 @@ in Section 4.1.2.1 of {{RFC6347}}.
 
 When receiving a datagram with the tls12_cid content type,
 the new MAC computation defined in {{mac}} MUST be used. When receiving a datagram
-with the RFC 6347-defined record format the MAC calculation defined in Section 4.1.2
+with the RFC 6347-defined record format, the MAC calculation defined in Section 4.1.2
 of {{RFC6347}} MUST be used.
 
 # Record Layer Extensions
@@ -192,7 +191,7 @@ three implications:
 
 - The CID field is present and contains one or more bytes.
 - The MAC calculation follows the process described in {{mac}}.
-- The true content type is inside the encryption envelope, as described
+- The real content type is inside the encryption envelope, as described
   below.
 
 Plaintext records are not impacted by this extension. Hence, the format
@@ -296,7 +295,7 @@ cid_length
 : 1 byte field indicating the length of the negotiated CID.
 
 length_of_DTLSInnerPlaintext
-: The length (in bytes) of the serialised DTLSInnerPlaintext (two-byte integer).
+: The length (in bytes) of the serialized DTLSInnerPlaintext (two-byte integer).
   The length MUST NOT exceed 2^14.
 
 seq_num_placeholder
@@ -378,9 +377,9 @@ data the following modification is made to the additional data calculation.
 # Peer Address Update {#peer-address-update}
 
 When a record with a CID is received that has a source address
-different than the one currently associated with the DTLS connection,
+different from the one currently associated with the DTLS connection,
 the receiver MUST NOT replace the address it uses for sending records
-to its peer with the source address specified in the received datagram
+to its peer with the source address specified in the received datagram,
 unless the following three conditions are met:
 
 - The received datagram has been cryptographically verified using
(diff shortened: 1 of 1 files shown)


ekr:
FWIW, I think some of these are gratuitous but they're not harmful.

PR #119, opened 2021-06-08
clarify "replace the address"

kaduk:
[just something I was thinking about in looking at John's ballot comment.  I'm not 100% sure whether this is actually needed or helpful.]

This refers to the default address to send stuff to, but should not
exclude one-offs that go elsewhere as deliberate probes.

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -380,7 +380,6 @@ data the following modification is made to the additional data calculation.
 When a record with a CID is received that has a source address
 different from the one currently associated with the DTLS connection,
 the receiver MUST NOT replace the address it uses for sending records
-(other than specific probing packets)
 to its peer with the source address specified in the received datagram,
 unless the following three conditions are met:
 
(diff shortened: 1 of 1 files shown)


boaks:
FMPOV, that may be amended "somehow" to the third "unless condition", though I consider the "probes" are part of that strategy.

> There is a strategy for ensuring that the new peer address is able to receive and process DTLS records.

kaduk:
Sure, "have a strategy for ensuring" could include sending the probes.  Which gets back to my initial disclaimer, I think :)

boaks:
I think the current wording was somehow based on the "implementation idea", that there is something as a  "verified address". That will only be replaced by a new one, only if that new one gets verified. Therefore my remark, that I would move the disclaimer somehow to that "strategy".

e.g.
unless ...

> There is a strategy for ensuring that the new peer address is able to receive and process DTLS records. No strategy is mandated by this specification but see note (*) below. (That strategy may use the different source address to send specific probing packets to.)

kaduk:
That seems worth trying out (new attempt pushed)

ekr:
I'm kind of negative on this one. I tend to think we've said too much already, given we decided to punt this.

kaduk:
> I'm kind of negative on this one. I tend to think we've said too much already, given we decided to punt this.

That's fair.  This started out as an internal thought "how can you probe a new address if you can't replace the address you send to?", but as Achim pointed out, we don't require that you have already completed a verification (before you send to the new address), just that you have a strategy for the verification, which is a weaker requirement.
Thanks for taking a look.

thomas-fossati:
We seem to think Section 6 is good as-is.  Since we have established that [dtls-rrc](https://github.com/tlswg/dtls-rrc) is the place where to deal with this matter, we agreed to close this PR

PR #120, opened 2021-06-08
Mention padding and ContentType encryption

kaduk:
As suggested by Éric Vyncke, mention that the new ciphertext record
format provides ContentType encryption and per-record padding in both
abstract and introduction.

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -69,6 +69,9 @@ is accomplished with the help of the 5-tuple. If the source IP address and/or
 source port changes during the lifetime of an ongoing DTLS session then the
 receiver will be unable to locate the correct security context.
 
+The new ciphertext record format with CID also provides content type encryption
+and record-layer padding.
+
 --- middle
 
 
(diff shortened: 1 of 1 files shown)


thomas-fossati (approved):

ekr (approved):
LGTM

draft-ietf-tls-dtls-connection-id.md
```diff
@@ -96,6 +99,11 @@ This document defines an extension to DTLS 1.2 to add a Connection ID (CID) to t
 DTLS record layer. The presence of the CID is negotiated via a DTLS
 extension.
 
+Adding a CID to the ciphertext record format presents an opportunity to make
+other changes to the record format.  In keeping with the best practices
+established by TLS 1.3, the type of the record is encrypted, and
+a mechanism provided for adding padding to obfuscate the plaintext length.
```
boaks, line 105: Being not a native english speaker, I feel, a "is" is missing? 

> and a mechanism is provided for adding padding to obfuscate the plaintext length.

thomas-fossati: Looks like a construction that in Italian grammar is called (literal translation) "elliptical proposition".  You can omit subject or predicate in a sentence if they are unambiguously implied by the context.

boaks: Thanks :-)

PR #122, opened 2021-08-06
Use wording similar to other length field.

boaks:
Fix question 5 rfced.

Signed-off-by: Achim Kraus <achim.kraus@bosch.io>


boaks:
Already integrated.

PR #123, opened 2021-08-06
Use same trailing ) position.

boaks:
Fix question 6 rfced.

Signed-off-by: Achim Kraus <achim.kraus@bosch.io>


boaks:
Already integrated.

PR #124, opened 2022-02-09
"DTLS was designed for securing connection-less transports, like UDP"

emanjon:
"(DTLS) {{RFC6347}} protocol was designed for securing connection-less transports, like UDP"

- I am not sure that "connection-less" is true. DTLS 1.0 (RFC4347) only mentions datagram protocols
and it was quickly used to protect data over connection-oriented datagram protocols like SCTP.
RFC6083bis (DTLS over SCTP) will likely rely heavily on DTLS Connection ID (at least that is
the current plan).

- Also DTLS in not really securing UDP, it secures the data sent over UDP.

### DIFF
--- a/draft-ietf-tls-dtls-connection-id.md
+++ b/draft-ietf-tls-dtls-connection-id.md
@@ -78,7 +78,7 @@ and record-layer padding.
 #  Introduction
 
 The Datagram Transport Layer Security (DTLS) {{RFC6347}} protocol was designed for
-securing connection-less transports, like UDP. DTLS, like TLS, starts
+securing application data sent over datagram protocols, like UDP. DTLS, like TLS, starts
 with a handshake, which can be computationally demanding (particularly
 when public key cryptography is used). After a successful handshake,
 symmetric key cryptography is used to apply data origin


boaks:
Not sure, if that change pays off (the draft has reached the editors queue).

It protects not only application data, it would protect alerts as well. (Even rehandshakes, but in my experience, that is not that common, usually just a full-epoch-0-handshake is used.)

About exchanging the phrase "connection-less transports" to "datagram protocols", yes, that would get closer to the terms in RFC6347.

[RFC6347](https://datatracker.ietf.org/doc/html/rfc6347)

> This document specifies version 1.2 of the Datagram Transport Layer
>    Security (DTLS) protocol.  The DTLS protocol provides communications
>    privacy for datagram protocols.

So maybe, just use the original phrase.

emanjon:
I think chaning "connection-less transports" to "datagram protocols" would be a good change in AUTH48.

boaks:
> The Datagram Transport Layer Security (DTLS) {{RFC6347}} protocol was designed for
securing connection-less transports, like UDP.

is now changed into 

> The Datagram Transport Layer Security (DTLS) protocol [[RFC6347](https://www.rfc-editor.org/rfc/authors/rfc9146.html#RFC6347)] was designed for securing data sent over datagram transports (e.g., UDP).

[rfc9146](https://www.rfc-editor.org/rfc/authors/rfc9146.html#section-1)

Hope, that is working for you?

boaks:
@korawit1188 

Just to mention: this work is already published as [RFC9146](https://www.rfc-editor.org/rfc/rfc9146.html).

